PR #10, opened 2023-08-26
Unify API wrapper

SleeplessOne1917:
This makes it so the API wrapper doesn't need to be defined twice.


src/api/mod.rs
```diff
@@ -97,32 +36,94 @@ where
...
+  }
 
   // Return the error response json as an error
   Response::de(&json).map_err(|_| LemmyAppError::APIError {
     error: json_deser_err(&json),
   })
 }
 
+#[macro_export]
+macro_rules! api_fn {
```
dessalines, line 111: I generally hate macros and meta-programming, because you lose compile-time checks, formatting, rust-analyzer support... all to save a few lines of code. Its not necessary here, verbose but checkable is preferred.

SleeplessOne1917: I removed the macro.

src/api/mod.rs
```diff
@@ -97,20 +36,69 @@ where
...
-        .await?
-    }
-    HttpType::Post => client.post(path).json(form).send().await?.text().await?,
-    HttpType::Put => client.put(path).json(form).send().await?.text().await?,
-  };
+  cfg_if! {
+    if #[cfg(feature = "ssr")] {
+     let client = reqwest::Client::new();
+
+     json = match type_ {
```
dessalines, line 44: Looks like you need a `let` here.

SleeplessOne1917: Clippy told me to get rid of the let.

dessalines (approved):

dessalines:
Nice, thx. I'll try to get the woodpecker stuff solved.

PR #19, opened 2023-09-30
Make leptos server proxy requests to lemmy server

SleeplessOne1917:
This started out as me trying to get the login page to work. However, I ended up needing to both change how the lemmy request client is setup as well as figure out how to proxy requests from the frontend server to the backend server. I'm submitting this for review and will continue working on the login page once these changes are merged.

PR #20, opened 2023-10-01
Authentication, Translations, Navigation updates

jim-taylor-business:
updates navigation with minimal styling (uses plain daisy/tailwind classes)

other changes included

- login and authentication fully working and integrated into lemmy API calls
- cookie handling (server and browser) for storing jwt token
- update leptos crates generally
- add i18n crate
- add translation files and translations for nav
- simple break-point layout
- pin lemmy API crate
- no styling on login as yet


jim-taylor-business:
@dessalines - we have come up with 2 different authentication approaches in these latest PRs

happy to go with whichever you think is best :)

SleeplessOne1917:
Some of the problems our PRs solve overlap and some don't. Some notes on your PR:
- I'm surprised you were able to make requests from the server functions to the API without setting up proxying as as done in this PR. Attempts to do that failed for me without setting up the actix server to handle requests to the /api/* route differently.
- Managing cookies on the client side seems like it has more errors that need to be handled than I would expect. I wonder if there's an easier way to to this.
- When running your changes locally, can you hit the API when using localhost:1237 as the host, e.g. localhost:1237/api/v3/post/list? Other lemmy UIs are going to need to have requests to the API go through for them to work.

SleeplessOne1917:
As a heads up, I'm likely going to close this PR without merging in the future after #19 gets merged. I have 3 reasons for this:

1. I think managing the cookies on the client side will be unnecessary, as cookies are already always sent with requests from the browser.
2. My PR should avoid creating more than one client on the server side. While my PR uses awc as the HTTP client library, [the reqwest docs also suggest creating only one client and reusing it](https://docs.rs/reqwest/latest/reqwest/struct.Client.html). If I understand correctly, the api wrapper currently creates a new client every time a request is made, which is something that should be avoided if possible.
3. I am going to experiment with a different approach to letting the browser side know that a user is logged in, as well as managing the state of the site more generally.

I will will wait a few days to see if @dessalines has any input on either of our PRs. If we hear nothing back from him after a few more days, I will merge my PR and get started on my attempt at session handling.

locales/en.json
```diff
@@ -0,0 +1,20 @@
+{
+  "nav_communities": "Communities",
```
dessalines, line 2: I don't think we should prefix these strings. We started doing that in jerboa and it turned out to be a huge mistake, with 4 different versions of `_communities` all being the same word.

src/api/mod.rs
```diff
@@ -32,73 +31,96 @@ pub async fn api_wrapper<Response, Form>(
...
+
+      let mut request_builder = match type_ {
+        HttpType::Get => gloo_net::http::Request::get(&build_fetch_query(route, form)),
+        HttpType::Post => gloo_net::http::Request::post(route),
+        HttpType::Put => gloo_net::http::Request::put(route),
+      };
+
+      match get_cookie_wrapper("jwt").await {
+        Ok(jwt) => {
+          if jwt.len() > 0 {
```
dessalines, line 79: This length check is probably pointless. If that cookie exists, then we should be fine.

src/api/mod.rs
```diff
@@ -115,3 +137,154 @@ fn build_fetch_query<T: Serialize>(path: &str, form: T) -> String {
...
+  };
+  use leptos::expect_context;
+  use leptos_actix::ResponseOptions;
+
+  let response = expect_context::<ResponseOptions>();
+
+  let mut cookie = Cookie::build(path, value).finish();
+  let mut now = OffsetDateTime::now_utc();
+  now += Duration::weeks(1);
+  cookie.set_expires(now);
```
dessalines, line 231: This should come from the cookie / auth response.

src/lib.rs
```diff
@@ -24,23 +31,40 @@ pub fn App() -> impl IntoView {
...
-        // <Route path="stories/:id" view=Story/>
-        // <Route path=":stories?" view=Stories/>
-        </Routes>
-      </main>
-      <BottomNav/>
+      <div class="flex flex-col h-screen">
+        // shows a progress bar while async data are loading
+        <RoutingProgress is_routing max_time=std::time::Duration::from_millis(250)/>
+        <TopNav/>
+        <main class="container mx-auto">
```
dessalines, line 42: I think it'd be better to let each Activity set its own container, and be fully configurable. Some might want to do some full-width things, who knows.

src/ui/components/common/nav.rs
```diff
@@ -1,61 +1,155 @@
...
+          </li>
+          <li>
+            <a href="//join-lemmy.org" class="text-md">
+              "join-lemmy.org"
+            </a>
+          </li>
+        </ul>
       </div>
-    </footer>
+    </nav>
```
dessalines, line 153: Looks good so far, nice!

src/lib.rs
```diff
@@ -24,23 +31,40 @@ pub fn App() -> impl IntoView {
...
+          <Routes>
+            <Route path="" view=HomeActivity ssr=SsrMode::Async/>
+            <Route path="home" view=HomeActivity ssr=SsrMode::Async/>
```
dessalines, line 45: These should all be able to work with out of order streaming tho, otherwise it'll be much slower.

dessalines:
@SleeplessOne1917 don't close this one, @jim-taylor-business fully got authentication working here, as well as a lot of other additions like translations and a fuller navbar. 

Work with them to fix some of the conflicts, and figure out the optimal way to do authentication.

src/api/mod.rs
```diff
@@ -115,3 +137,154 @@ fn build_fetch_query<T: Serialize>(path: &str, form: T) -> String {
   let form_str = serde_urlencoded::to_string(&form).unwrap_or(path.to_string());
   format!("{path}?{form_str}")
 }
+
+#[cfg(not(feature = "ssr"))]
```
SleeplessOne1917, line 141: I don't think we have to worry about cookies on the browser side.

src/api/mod.rs
```diff
@@ -115,3 +137,154 @@ fn build_fetch_query<T: Serialize>(path: &str, form: T) -> String {
...
+  use leptos_actix::ResponseOptions;
+
+  let response = expect_context::<ResponseOptions>();
+
+  let mut cookie = Cookie::build(path, "").finish();
+  let mut now = OffsetDateTime::now_utc();
+  now += Duration::weeks(-1);
+  cookie.set_expires(now);
+  cookie.set_path("/");
+
```
SleeplessOne1917, line 261: Does actix cookies not have removing cookies as part of their API?

src/api/mod.rs
```diff
@@ -115,3 +137,154 @@ fn build_fetch_query<T: Serialize>(path: &str, form: T) -> String {
...
+      "".to_string()
+    }
+  })
+  .await
+  .map_err(|e| LemmyAppError::APIError {
+    error: e.to_string(),
+  })?;
+
+  Ok(cookie_value)
+}
```
SleeplessOne1917, line 290: There is an [actix-session crate](https://docs.rs/actix-session/latest/actix_session/) that should simplify handling cookies. It will allow you to set up a middleware that lets you extract cookies in a form that's easier to user that what is being done currently.

src/ui/components/comment/comment_nodes.rs
```diff
@@ -9,7 +9,7 @@ pub fn CommentNodes(comments: MaybeSignal<Vec<CommentView>>) -> impl IntoView {
       <For
         each=comments
         key=|cv| cv.comment.id
-        view=move |cv| {
+        children=move |cv| {
             view! {
               <li>
                 <CommentNode comment_view=cv.into()/>
```
SleeplessOne1917, line 15: When you merge main, keep the way main is doing it. Leptos 0.5.0 let's you pass children as children in the markup instead of passing a closure to a component prop.

src/ui/components/common/nav.rs
```diff
@@ -1,61 +1,155 @@
...
+pub async fn logout_form_fn() -> Result<(), ServerFnError> {
+  use crate::api::set_cookie_wrapper;
+  use leptos_actix::redirect;
+
+  redirect("/");
+
+  match set_cookie_wrapper("jwt", "").await {
+    Ok(o) => Ok(o),
+    Err(e) => Err(ServerFnError::ServerError(e.to_string())),
+  }
```
SleeplessOne1917, line 16: I would assume the `remove_cookie_wrapper` that was made in the other file is for this purpose. This all may be moot if we go with actix-session for cookie management.

src/ui/components/common/nav.rs
```diff
@@ -1,61 +1,155 @@
...
+              {t!(i18n, nav_create_post)}
+            </A>
+          </li>
+          <li>
+            <A href="/create_community" class="text-md">
+              {t!(i18n, nav_create_community)}
+            </A>
+          </li>
+          <li>
+            <A href="//join-lemmy.org/donate">
```
SleeplessOne1917, line 50: Starts with 2 slashes. I presume this is unintentional.

src/ui/components/common/nav.rs
```diff
@@ -1,61 +1,155 @@
...
+          <li>
+            <a href="//github.com/LemmyNet/lemmy-ui-leptos/releases" class="text-md">
+              "f/e from env"
+            </a>
+          </li>
+          <li>
+            <a href="//github.com/LemmyNet/lemmy/releases" class="text-md">
+              "b/e from env"
+            </a>
+          </li>
```
SleeplessOne1917, line 125: Are the double-slashes at the beginning of the hrefs intentional?

locales/en.json
jim-taylor-business: i was concerned that the context of the word/phrase can affect the meaning and therefore the words being used, specially in other languages.

it sounds like your experience with jerboa indicates that is not a concern so i will remove the context

src/api/mod.rs
jim-taylor-business: agreed

src/lib.rs
jim-taylor-business: ok

src/ui/components/common/nav.rs
jim-taylor-business: i'm being ultra vanilla with styling until we decide on a direction. the less code the better maybe at this point :)

src/api/mod.rs
jim-taylor-business: i can't find it explicitly, can you?

i think this is a standard way of removing cookies but i'm open to alternatives

src/ui/components/comment/comment_nodes.rs
jim-taylor-business: sure, this was just a quick fix as i wasn't focused on this area

src/api/mod.rs
jim-taylor-business: this was created for completeness as i was developing the whole solution. 

i thought it could be useful in the future and also the code implied that there should be an equivalent function when compiling in "ssr" and "hydrate" modes.

maybe we can refactor this away later as part of a cookie oriented refactoring?

src/api/mod.rs
jim-taylor-business: maybe we can refactor this away later as part of a cookie oriented refactoring?

src/ui/components/common/nav.rs
jim-taylor-business: without the double slash it get interpreted as a relative URL.

src/ui/components/common/nav.rs
jim-taylor-business: without the double slash it get interpreted as a relative URL.

src/ui/components/common/nav.rs
SleeplessOne1917: I see. I think external urls use the lowercase a tag.

src/api/mod.rs
dessalines: IMO its perfectly fine to leave this cookie stuff in here, we can always remove it at a later date.

locales/en.json
SleeplessOne1917, line 1: Should translations be in a separate repo and included as a submodule like in the original UI?

SleeplessOne1917 (approved):

locales/en.json
jim-taylor-business: the scope of this PR was simply to prove the concept.

to avoid the scope of this PR creeping i think this can and should be addressed in a future PR and maybe discussed in seperate ticket

src/api/mod.rs
jim-taylor-business: do you mean the expiry date?

sorry i don't think i see one in the login response.

src/lib.rs
jim-taylor-business: i think the out of order streaming requires javascript to work which negates the benefits of a fully SSR pages

can we change this when we have a full understanding of which pages and how much of those pages needs to be rendered on the server?

dessalines:
There's a `./format.sh` script in the top level dir, I'll run it for ya here.

dessalines (approved):

PR #21, opened 2023-10-06
Changes to allow Login and Logout to function better in SSR only mode

jim-taylor-business:
adjusts login and logout behaviour depending on whether javascript is enabled.

you may have noticed that there are signals set up now to make the login/logout cause updates around the app using a provide_context signal.

other changes are non-functional


jim-taylor-business:
would love to get this PE in as well @dessalines as it completes my recent work

thanks for your help

SleeplessOne1917 (approved):

PR #25, opened 2023-10-10
Handle cookies and login state

### LINKED ISSUE #22: Issue with new lemmy_client
@SleeplessOne1917 @dessalines 

i was unable to get the new lemmy client to make any successful request apart from the home page content (list_posts) on the server side only.

i noticed the home page content has been commented out completely so i assume this was a known issue.

i have left all the code there and only added some prototype code to add the authentication header.

SleeplessOne1917:
This fixes #22 (hopefully). It makes cookie management a little easier.

In addition, I'm using leptos_query to handle fetching the site state to track if someone is logged in. It should alow us to use a resource everywhere around the app without unnecessary refetches or more complicated to use contexts.

Some of the markup in the navbar was removed. I will put this back in before I merge, I just chopped a lot of stuff out to make it easier to manage showing login/logout.


dessalines:
Sounds good, just ping me when you're ready.

src/ui/components/common/nav.rs
```diff
@@ -1,213 +1,151 @@
...
+                    <li>
+                      <A href="/settings">{t!(i18n, nav.settings)}</A>
+                    </li>
+                    <li>
+                      <hr/>
+                    </li>
+                    <li>
+                      <ActionForm action=logout_action>
+                        <button type="submit">{t!(i18n, nav.logout)}</button>
+                      </ActionForm>
```
SleeplessOne1917, line 121: This button won't submit the form at all when in this details, so logging out is temporarily broken. I will clean this up in a future PR.

SleeplessOne1917:
@dessalines @jim-taylor-business While I'll be cleaning and refactoring this a bit more, the lion's share of the work is done.

Cargo.toml
```diff
@@ -28,25 +30,27 @@ gloo-net = { version = "0.4", features = ["http"] }
 log = "0.4"
 cfg-if = "1.0"
 tracing = "0.1"
-reqwest = { version = "0.11", features = ["json"] }
 getrandom = { version = "0.2", features = ["js"] }
-lemmy_api_common = { tag = "0.19.0-rc.1", git = "https://github.com/LemmyNet/lemmy.git", default-features = false }
+lemmy_api_common = { git = "https://github.com/LemmyNet/lemmy.git", branch = "main", default-features = false }
```
jim-taylor-business, line 34: the version tag was good for not getting every single update from lemmy which is nice for stability. was there something broken or missing from lemmy for you?

Cargo.toml
```diff
@@ -28,25 +30,27 @@ gloo-net = { version = "0.4", features = ["http"] }
...
 
+wasm-cookies = "0.2.1"
```
jim-taylor-business, line 36: i see you made this more specific than before, this would have been picked up by a cargo update of version "0.2"

is there a reason for being more specific?

src/config.rs
```diff
@@ -1 +1 @@
-pub const TEST_HOST: &str = "voyager.lemmy.ml";
+pub const TEST_HOST: &str = "localhost:8536";
```
jim-taylor-business, line 1: does this work with voyager? i am biassed as that is what i am currently using for testing :smile: 

would it be possible to specify protocol in this url (https) as i think that is required for this to work with voyager.

Cargo.toml
SleeplessOne1917: I was getting deserialization errors because the local_user on the site response had a field on it returned from the server that the current version of lemmy_api_common can't handle.

Cargo.toml
SleeplessOne1917: No. I can put it back to 0.2.

src/config.rs
SleeplessOne1917: I've been preferring localhost since I can look at what gets logged on the server when I come across particularly annoying bugs. As for protocol, I believe I have a method/some logic somewhere that handles that.

src/ui/components/common/nav.rs
```diff
@@ -1,213 +1,151 @@
...
   let i18n = use_i18n();
-
   view! {
     <nav class="container navbar mx-auto">
       <div class="navbar-start"></div>
       <div class="navbar-end ">
         <ul class="menu menu-horizontal flex-nowrap">
           <li>
-            <a href="//github.com/LemmyNet/lemmy-ui-leptos/releases" class="text-md">
+            <a href="github.com/LemmyNet/lemmy-ui-leptos/releases" class="text-md">
```
jim-taylor-business, line 143: i think the double slash is needed for the url to be absolute.

src/ui/components/home/home_activity.rs
```diff
@@ -64,27 +55,27 @@ pub fn HomeActivity() -> impl IntoView {
...
+    // }
+    // Some(res) => {
+    // view! {
+    // <div>
+    // <PostListings posts=res.posts.into()/>
+    // </div>
+    // }
+    // }
+    // })
+    // }}
```
jim-taylor-business, line 76: it would be nice to see this working with the new reactuve auth system

jim-taylor-business:
it looks really cool so far but it would be nice to see the home page and logout working as part of the whole proof of concept

Cargo.toml
jim-taylor-business: ah i see. i guess that's because your local dev server is using the very latest version so you need them to be in sync.

i personally quite like the server api to be stable while i'm working against it.

i guess we can handle this locally.

src/config.rs
jim-taylor-business: so i will need to change the code rather than configuration to change the protocol it uses?

jim-taylor-business:
testing the login on my local machine works against Voyager with the URL and protocol changes :+1: 

the navbar does not react to the login however and show the user menu for example.

src/lib.rs
```diff
@@ -25,50 +27,36 @@ leptos_i18n::load_locales!();
...
-    // adding `set_is_routing` causes the router to wait for async data to load on new pages
     <Router set_is_routing>
-      <div class="flex flex-col h-screen">
-        <RoutingProgress is_routing max_time=std::time::Duration::from_millis(250)/>
-        <TopNav/>
-        <Routes>
-          <Route path="" view=HomeActivity ssr=SsrMode::Async/>
-          <Route path="home" view=HomeActivity ssr=SsrMode::Async/>
+      <Routes>
+        <Route path="/" view=move || view! { <Layout is_routing/> } ssr=SsrMode::PartiallyBlocked>
```
jim-taylor-business, line 37: this is cool! i think `SsrMode::PartiallyBlocked` . i think this might be new since i last looked

seems like the perfect balance for SSR performance. looking forward to see how this works.

jim-taylor-business:
login is working perfectly with JS disabled as well :+1:

jim-taylor-business:
i'm getting this dumped in the console although it does not seem to affects the functionality of the site at all, as it is

jim-taylor-business:
by the way, as i alluded to in my PR, getting all this working perfectly has been a real challenge. i've struggled a lot.

i've had generic errors on both browser and server that have frustrated me at every turn.

i'm sure once we find the pattern that just works :tm: we'll be able to copy it across the site and things will be awesome.

src/ui/components/common/nav.rs
SleeplessOne1917: You're right. Added them back in.

src/config.rs
SleeplessOne1917: You can pass in a LEMMY_UI_HTTPS environment variable to have it use HTTPS.

SleeplessOne1917:
@jim-taylor-business The panic that's not affecting anything seems to be from leptos_query. I opened an issue about it on its repo.

jim-taylor-business:
would you be ok with me merging your code into mine manually - not a git merge but something i will do by hand @SleeplessOne1917 ?

Cargo.toml
dessalines: I do agree we should try to stay versioned, but ya I'll try to start another back-end beta deploy.

Cargo.toml
dessalines: Also be sure to have `cargo edit` installed on your systems, and use `cargo upgrade`, which will overwrite a lot of these anyway.

SleeplessOne1917:
@jim-taylor-business that sounds more difficult and error prone than git merging.

jim-taylor-business:
i will be very careful, i promise you

SleeplessOne1917:
This PR touches a lot of files. I'd rather merge this into main first instead of making an even bigger PR with more changes.

jim-taylor-business (approved):

jim-taylor-business:
when you're ready @SleeplessOne1917 go for it

PR #28, opened 2023-10-12
Home improvements

jim-taylor-business:
i have definitely been wrestling with signal flows and i have been using create_effects as a fix which does not "feel" correct. i know we are using a lot of "nightly" so maybe that's why, but probably my ignorance is more to blame. obviously i will be coming back to this over time and will nail it.

nice things:

- themes!
- fully responsive to login status

known issues:

- janky design, yup it's a little janky
- missing trending communities and instance summary
- no thumbnail previews
- community filter non functional
- content type filter non functional
- content sort non functional
- no translations
- the prev and next navigation cannot currently work on server side and will be rewritten


SleeplessOne1917:
1. Could you please include some screenshots?
2. I'd recommend holding off merging this until #25 is merged since it makes session handling a bit simpler, as well as improving the client for API calls. I could see it creating some merge conflicts for this PR.

jim-taylor-business:
this PR is building on the current system, it does not change or extend it in any way, so it should not create further conflicts. even theming was very light to add on top of what we have.

i guess what you're doing is a larger refactor and not something that can be swapped for what exists without breaking the current structure.

i'm keen to get both merged so i will focus on your PR

jim-taylor-business:
as i said in the Home page ticket recently the only design i'm going for is pure Daisy to the point of 100% functionality.

i wanted to get this imperfect version in early because the home page is a big chunk and i wanted to phase it rather than wait until absolutely everything is finished.

it ain't perty but it's not too bad :smile: 

[Screencast from 2023-10-13 07-49-56.webm](https://github.com/LemmyNet/lemmy-ui-leptos/assets/7439891/9ff502b2-ef7f-47aa-9d2f-0c16efa22992)

src/ui/components/home/home_activity.rs
```diff
@@ -14,56 +17,80 @@ pub async fn list_posts(form: &GetPosts) -> Result<GetPostsResponse, LemmyAppErr
...
+
+  let error = create_rw_signal::<Option<String>>(None);
+
+  let authenticated = expect_context::<RwSignal<bool>>();
+
+  let next_page_cursor = create_rw_signal::<Option<PaginationCursor>>(None);
+  let page_cursor = create_rw_signal::<Option<PaginationCursor>>(None);
 
-  let authenticated = use_context::<RwSignal<bool>>().unwrap_or(create_rw_signal(false));
+  let prev_cursor_stack = create_rw_signal::<Vec<Option<PaginationCursor>>>(vec![]);
```
dessalines, line 29: Having just done this for lemmyUI, unfortunately you pretty much need to rely on the browser back button, because keeping a stack won't work. IE hitting the `previous` button actually might go forward a page, since that might have been the last page you visited.

dessalines (approved):
This is a good start, I'm not against merging. Once things get a bit more developed, we can make these PRs more page and feature specific.

src/ui/components/home/home_activity.rs
jim-taylor-business: :laughing: yeah i have tried to replicate a back button there, browser back would be much easier although doesn't work without JS?

dessalines: I'd have to test but I don't see why the leptos router wouldn't work without js.

PR #30, opened 2023-10-14
Try to fix clippy in CI.

dessalines:


SleeplessOne1917 (approved):
I figure this is the kind of thing we won't know works until we trigger a build?

dessalines:
The build passed.

PR #31, opened 2023-10-15
Themes, home functionality

jim-taylor-business:
nice things:

- themes
- favicon
- voting, saving, reporting

known issues:

- post state should be raised from PostListing so that post data managed in Activity component, this is so that more sophisticated interactions with the Post list and interactions (errors, feedback) can be enabled  
- fully responsive to login status
- SSR, still issues with SSR only flow and error handling
- janky design, yup it's a little janky
- missing trending communities and instance summary
- no thumbnail previews
- community filter non functional
- content type filter non functional
- content sort non functional
- no translations
- the prev and next navigation cannot currently work on server side and will be rewritten


src/api/mod.rs
SleeplessOne1917, line 1: This file isn't needed anymore because the code in `lemmy_client.rs` takes care of it.

src/ui/components/common/nav.rs
```diff
@@ -142,19 +169,29 @@ pub fn TopNav() -> impl IntoView {
 #[component]
 pub fn BottomNav() -> impl IntoView {
   let i18n = use_i18n();
+
+  let QueryResult { data, refetch, .. } = use_site_state();
+
+  let instance_api_version =
+    Signal::derive(move || data().map_or_else(|| None, |res| Some(res.ok()?.version)));
```
SleeplessOne1917, line 176: This might be simpler as:
```suggestion
    Signal::derive(move || data().map(|res| res.ok().map(|res| res.version)));
```

src/ui/components/login/login_form.rs
```diff
@@ -28,6 +28,7 @@ pub async fn login(username_or_email: String, password: String) -> Result<(), Se
       session.insert("jwt", jwt.into_inner())?;
     }
 
+    redirect("/");
```
SleeplessOne1917, line 31: This is taken care of in the isomorphic effect. I can't do it in this server function because I need to refetch the site response, which seems to only be doable in the component.

src/ui/components/post/post_listing.rs
```diff
@@ -1,16 +1,306 @@
...
+    }
+  });
+
   view! {
-    <A href=link class="link">
-      {pv.creator.name}
-      -
-      {pv.post.name}
-    </A>
+    <tr>
```
SleeplessOne1917, line 163: I don't think table rows are the way to handle this, and they don't make sense outside of a table table tag and thead/tbody tag.

jim-taylor-business:
thanks @SleeplessOne1917 this is still a Draft PR.

just wanted to let people know that i'm working on the merge

src/ui/components/login/login_form.rs
jim-taylor-business: is required for SSR only login, the effect is not triggered

dessalines (commented):
This also appears to break navigation, I can't exactly tell why.

src/config.rs
```diff
@@ -1 +1,2 @@
-pub const TEST_HOST: &str = "localhost:8536";
+pub const TEST_HOST: &str = "voyager.lemmy.ml";
```
SleeplessOne1917, line 1: Why voyager? Still having trouble getting it working locally?

src/layout.rs
```diff
@@ -5,24 +5,22 @@ use leptos_router::{Outlet, RoutingProgress};
 
 #[component]
 pub fn Layout(is_routing: ReadSignal<bool>) -> impl IntoView {
+  let ui_theme = expect_context::<RwSignal<String>>();
```
SleeplessOne1917, line 8: I imagine this will be part of the site response down the line.

src/layout.rs
```diff
@@ -5,24 +5,22 @@ use leptos_router::{Outlet, RoutingProgress};
...
-        <Meta name="viewport" content="viewport-fit=cover"/>
-        <Script src="//cdn.jsdelivr.net/npm/eruda"/>
-        <Script>eruda.init();</Script>
-        <Title text="Brand from env"/>
+    <RoutingProgress is_routing max_time=std::time::Duration::from_millis(250)/>
+    <Stylesheet id="leptos" href="/pkg/lemmy-ui-leptos.css"/>
+    <Link rel="shortcut icon" type_="image/ico" href="/favicon.svg"/>
+    <Meta name="description" content="Lemmy-UI-Leptos."/>
+    <Meta name="viewport" content="viewport-fit=cover"/>
+    // <Script src="//cdn.jsdelivr.net/npm/eruda"/>
```
SleeplessOne1917, line 16: @dessalines what is eruda? I saw it's included in the old UI and the join-lemmy repos when in debug mode, but I don't know what it adds that isn't already handled by browser devtools.

src/queries/site_state_query.rs
```diff
@@ -2,9 +2,9 @@ use lemmy_api_common::site::GetSiteResponse;
 use leptos::*;
 use leptos_query::{use_query, QueryOptions, QueryResult, RefetchFn, ResourceOption};
 
-#[server(GetSiteResource, "serverfn", "GetJson")]
+#[server(GetSiteResource, "/serverfn", "GetJson")]
```
SleeplessOne1917, line 5: The slash isn't needed.

src/ui/components/post/post_listing.rs
```diff
@@ -1,16 +1,288 @@
...
-  let link = format!("post/{}", pv.post.id);
+pub fn PostListing(
+  post_view: MaybeSignal<PostView>,
+  error: RwSignal<Option<String>>,
+) -> impl IntoView {
+  let authenticated = expect_context::<RwSignal<bool>>();
+
+  let post_view = create_rw_signal(post_view().clone());
+
+  let vote_action = create_server_action::<VotePostFn>();
```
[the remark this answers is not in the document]
SleeplessOne1917: The use of server actions with ActionForms means I think it will make more sense to keep the actions close to where they are used. I'm pretty sure regular functions can't be used for actions. However, defining the server functions in a separate file could help keep things tidy since so many are used here.

src/ui/components/post/post_listings.rs
```diff
@@ -3,15 +3,15 @@ use lemmy_api_common::lemmy_db_views::structs::PostView;
...
 
 #[component]
-pub fn PostListings(posts: MaybeSignal<Vec<PostView>>) -> impl IntoView {
+pub fn PostListings(
+  posts: MaybeSignal<Vec<PostView>>,
+  error: RwSignal<Option<String>>,
+) -> impl IntoView {
   view! {
-    <ul>
+    <table class="table">
```
SleeplessOne1917, line 11: Why use table?

src/queries/site_state_query.rs
jim-taylor-business: without the slash this gets interpreted as relative (try it on the PostActivity page) so we need the slash

src/lib.rs
```diff
@@ -29,30 +29,36 @@ pub fn App() -> impl IntoView {
   provide_i18n_context();
   provide_query_client();
 
+  let ui_theme = create_rw_signal::<String>(String::from("retro"));
+  provide_context(ui_theme);
+
+  let authenticated = create_rw_signal::<bool>(false);
```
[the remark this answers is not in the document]
jim-taylor-business: this or something like this will be needed to send a signal to the rest of the page that we are logged in when the pages is Hydrated. so we do need this

src/layout.rs
jim-taylor-business: yeah, good point, an instance could have default theme in future 

this signal will carry that theme identifier to the part of the page where it needs to be set

Cargo.toml
```diff
@@ -31,15 +39,13 @@ log = "0.4"
 cfg-if = "1.0"
 tracing = "0.1"
 getrandom = { version = "0.2", features = ["js"] }
-lemmy_api_common = { git = "https://github.com/LemmyNet/lemmy.git", branch = "main", default-features = false }
+lemmy_api_common = { tag = "0.19.0-rc.1", git = "https://github.com/LemmyNet/lemmy.git", default-features = false }
```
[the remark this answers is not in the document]
jim-taylor-business: the cargo definition only allows either branch or tag, not both

src/layout.rs
```diff
@@ -5,24 +5,22 @@ use leptos_router::{Outlet, RoutingProgress};
...
-      <ErrorBoundary fallback=|_| view! { Error! }>
-        <RoutingProgress is_routing max_time=std::time::Duration::from_millis(250)/>
-        <Stylesheet id="leptos" href="/pkg/lemmy-ui-leptos.css"/>
-        <Link rel="shortcut icon" type_="image/ico" href="/favicon.ico"/>
-        <Meta name="description" content="Lemmy-UI-Leptos."/>
-        <Meta name="viewport" content="viewport-fit=cover"/>
-        <Script src="//cdn.jsdelivr.net/npm/eruda"/>
-        <Script>eruda.init();</Script>
-        <Title text="Brand from env"/>
+    <RoutingProgress is_routing max_time=std::time::Duration::from_millis(250)/>
```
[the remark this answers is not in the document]
jim-taylor-business: the SSR Mode of the page (Route) decides how many of the <Suspense> areas that get awaited or rendered at all.

if we default to PartiallyBlocked we have very limited (1?) async blocks that will be rendered. having one at the top level does not actually bound any specific async anyway so i think it is not functional.

testing things with javascript disabled you get a very different view of the site. SSR only mode is actually very limited in its abilities.

i guess having a catch all error boundary at the top level is somewhat useful for the user in the event of a completely unhandled app error, although it would be nice to catch everything at a much lower level and still provide some UI.

src/ui/components/post/post_listing.rs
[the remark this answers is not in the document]
jim-taylor-business: at this stage i can see both sides on this. at the moment i agree with @SleeplessOne1917 although i can see how it would be useful or more efficient to manage the component state in the parent.

maybe we can address this later as this would be a significant amount more work @dessalines?

src/ui/components/post/post_listings.rs
SleeplessOne1917: I disagree that this is tabular data, but the markup can be changed in a later PR.

src/lib.rs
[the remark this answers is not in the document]
SleeplessOne1917: The site is a blocking resource. It will have already been fetched by the time the HTML is returned to the browser.

Cargo.toml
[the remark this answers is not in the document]
SleeplessOne1917: Then use the branch.

src/lib.rs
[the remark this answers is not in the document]
jim-taylor-business: if the site query can be used as a signal then that should work. 

will remove this as it can't work in SSR only mode anyway.

src/ui/components/post/post_listing.rs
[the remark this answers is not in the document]
dessalines: That's okay I spose, but it seems like the `state` / createEffect functions could be moved up a level, to make things like the `PostListing` stateless.

dessalines (commented):
Navigation is still broken, but @SleeplessOne1917 feel free to merge if you think its in a good enough state. Since this is in rapid early dev, its pry better to merge things that are half broken, to make collaboration easier.

src/ui/components/post/post_listing.rs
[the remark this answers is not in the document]
jim-taylor-business: ok great, so we can acknowledge this as a known design issue and approve the PR. i will add another known issue to the PR 

this PR is far from perfect (see all the know issues) but i would like to merge it so that we can all move on without "treading on each others toes" as we have been doing a little bit in these early PRs

jim-taylor-business:
> Navigation is still broken, but @SleeplessOne1917 feel free to merge if you think its in a good enough state. Since this is in rapid early dev, its pry better to merge things that are half broken, to make collaboration easier.

yes there is still issues with almost everything in this PR but it is just a stepping stone to merging and stabilising the base functionality of the site and moving on to new features without "stepping on each-others toes" as has been happening a bit in previous PRs

SleeplessOne1917 (approved):

SleeplessOne1917:
@jim-taylor-business You should be good to open a PR with your changes. After that, I'll be doing some tweaking myself. Hopefully we can get this project to a point where it's easier to work without stepping on eachother's toes.

PR #36, opened 2023-10-20
Handle module file layout differently

SleeplessOne1917:
Using named files for modules. Could make it easier to find modules than many files named mod.rs.


jim-taylor-business:
this is a rust standard, there is no references to discussion or design, 

this is very disruptive to current work

maybe we can focus on fixing existing issues and old features that have been disabled since the client refactoring (PostActivity).

in short i see no need for this.

PR #37, opened 2023-10-20
Fix direct communication from browser to server when spcified in ENV variable

jim-taylor-business:
updates:

- if you're using a local Lemmy instance and the internal proxy this update will continue to work without configuration change. proxy code logic and functionality untouched
- feature bypass_internal_proxy enables direct communication in the browser
- environment vars LEMMY_UI_LEMMY_INTERNAL_HOST ("host:port") and LEMMY_UI_HTTPS ("true" | "false") will be used on server and baked into browser WASM on compile to connect to Lemmy instance. defaults are still in config.rs

known issues:

- Nav logic completely disabled because it hangs the page and/or errors and crashes the WASM
- HomeActivity paging browser only as i cannot find a way to convert PaginationCursor to/from string 

clippy is crashing again and blocking PR. clippy doesn't work locally either


Cargo.toml
```diff
@@ -64,7 +64,7 @@ locales = ["en", "fr"]
 namespaces = ["nav"]
 
 [features]
-default = ["csr"]
+default = ["ssr"]
```
SleeplessOne1917, line 67: Why make SSR the default?

jim-taylor-business: csr feature is not a target for us now. ssr is findamental

dessalines:
I think I got clippy fixed now.

src/server/api_service.rs
```diff
@@ -1,43 +1,41 @@
...
+pub async fn route_to_api(
+  request: HttpRequest,
+  client: web::Data<Client>,
+  payload: web::Payload,
+  uri: Uri,
+) -> Result<HttpResponse, ProxyError> {
+  let mut parts = uri.into_parts();
+  parts.authority = Some(get_host().try_into().map_err(map_uri_err::<InvalidUri>)?);
+  parts.scheme = Some(
+    format!("http{}", if get_https() == "true" { "s" } else { "" })
```
dessalines, line 23: GJ making these external requests automatically use the browser url. Much easier than having to define another env.

dessalines (approved):

src/server/api_service.rs
jim-taylor-business: oh i think that logic was already in there :smile: .. but it reminds me of the case when you're using the proxy it should adopt the protocol the page is using, another tiny PR incoming!

would be awesome if PaginationCursor can be easily to/from string, then i can page the list of posts from the server side :+1:

jim-taylor-business:
> I think I got clippy fixed now.

awesome, was it easy? i am unable to run it locally :disappointed:

dessalines:
Ya, just needed to change the target.

PR #38, opened 2023-12-07
Try using cargo binstall for CI tasks.

dessalines:


Cargo.toml
```diff
@@ -32,13 +32,10 @@ leptos_i18n = { version = "0.2", features = ["nightly"] }
...
 
 serde = { version = "1", features = ["derive"] }
-serde_json = "1"
 web-sys = { version = "0.3", features = ["AbortController", "AbortSignal"] }
 gloo-net = { version = "0.4", features = ["http"] }
-log = "0.4"
 cfg-if = "1.0"
-tracing = "0.1"
-getrandom = { version = "0.2", features = ["js"] }
+#getrandom = { version = "0.2", features = ["js"] }
```
SleeplessOne1917, line 38: Why commented instead of removed?

.woodpecker.yml
```diff
@@ -22,10 +25,24 @@ steps:
 
   leptos_fmt:
     image: *rust_image
+    group: format
     commands:
-      - cargo install leptosfmt
+      - wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz
```
SleeplessOne1917, line 30: Why musl? Also, is there a way to avoid needing to download and extract binstall each time it is used?

Cargo.toml
dessalines: Fixed

.woodpecker.yml
```diff
@@ -1,19 +1,23 @@
 variables:
-  - &rust_image "rustlang/rust:nightly-slim"
+  - &rust_image "rustlang/rust:nightly"
+  - &install_binstall "wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz && tar -xvf cargo-binstall-x86_64-unknown-linux-musl.tgz && cp cargo-binstall /usr/local/cargo/bin"
```
dessalines, line 3: I've combined these up here now.

SleeplessOne1917: Nice! Is there a way to use the same downloaded b-install across the different steps instead of downloading it repeatedly? Downloading it more than once seems like it will add unnecessary time to the build.

dessalines: Not that I know of, because I'm not exactly sure how to properly do resource sharing in woodpecker. But they all finish within 15s anyway, and its done in parallel: https://woodpecker.join-lemmy.org/repos/140/pipeline/303/3

SleeplessOne1917 (approved):

PR #41, opened 2024-01-24
Use SVG sprites for icons

SleeplessOne1917:


src/ui/components/common/icon.rs
```diff
@@ -0,0 +1,53 @@
...
+      IconType::Search => "search",
+      IconType::Upvote => "upvote",
+      IconType::VerticalDots => "vertical-dots",
+    }
+  }
+}
+
+#[component]
+pub fn Icon(
+  #[prop(into)] icon: MaybeSignal<IconType>,
```
dessalines, line 43: This doesn't need to be a signal I don't think. Doesn't need state / can be made at compile time.

dessalines (approved):
Feel free to merge whenev

src/ui/components/common/icon.rs
SleeplessOne1917: I could see a scenario where there is one `Icon` component that has its icon type change. `MaybeSignal` accepts both signal and non-signal attributes.

SleeplessOne1917:
Is there a way to bypass CI for merging? Woodpecker always seems to act strange with this project.

PR #42, opened 2024-01-24
Update daisy ui to 4.6.0 and use pnpm for package manager

SleeplessOne1917:
Daisy UI had a major version update. This PR migrates to that. Fortunately, none of the breaking changes affect us.

Also replaced npm with pnpm, which is generally more efficient than npm.


dessalines (commented):
Nice, did some reading on pnpm vs yarn, looks good to me.

.woodpecker.yml
```diff
@@ -57,10 +57,11 @@ steps:
       - *install_binstall
       - cargo-binstall -y cargo-leptos
       - apt-get update && apt-get -y install libssl-dev pkg-config
-      - curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash
+      - curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
       - . /root/.profile
-      - nvm install node
-      - npm i
+      - nvm install lts
```
dessalines, line 62: This looks like it needs to be `nvm install --lts` .

dessalines (approved):
Feel free to merge when you get it passing CI.

PR #45, opened 2024-02-08
Lemmy client

SleeplessOne1917:
Replace the client that is currently used with the one from the rust Lemmy API wrapper I recently published.


src/server/api_service.rs
SleeplessOne1917, line 1: I removed this on the assumption that the address being targeted is where you are running the backend server, e.g. API requests made from the browser to the api route on localhost will go straight to the lemmy server at localhost, not voyager.lemmy.ml or some other domain. If this proxying functionality is needed/desired, I can add this back in.

SleeplessOne1917 (commented):
I've been making a lot of progress on this recently, and have forced through some changes (like the upgrade to leptos 0.6). I plan on bypassing approvals if this doesn't get any feedback in the next day or so. I'm starting to pick up steam in this project and want to really get in to making everything.

I also encourage you to test this locally on your machine while also running the Lemmy backend locally. I've found that it works very smoothly with no errors or warnings, but it's always good to double check.

If you're having issues loading the site, check if the existing frontend is broken as well. I've noticed when running the local lemmy docker compose that I have to delete my volumes and re-run docker_update.sh to get things working smoothly again.

src/ui/components/post/post_listing.rs
SleeplessOne1917, line 1: I plan on cleaning up the post listings once this PR is merged.

src/server/api_service.rs
dessalines: Seems fine. We can always add back in later.

src/ui/components/login/login_form.rs
```diff
@@ -24,7 +26,10 @@ pub async fn login(username_or_email: String, password: String) -> Result<(), Se
     totp_2fa_token: None,
   };
 
-  let LoginResponse { jwt, .. } = client.login(req).await?;
+  let LoginResponse { jwt, .. } = client
```
dessalines, line 29: Avoid these javascripty destructurings, they're not really necessary in rust, and are really confusing to look at.

IE `let jwt = client...map_err...?.jwt`

dessalines (approved):
I was able to log in with this, so nice job on the client.

We might want to consider adding in cargo-husky or something similar, in order to run leptosfmt on every commit. Also for prettier as well.

src/ui/components/login/login_form.rs
SleeplessOne1917: I can see avoiding destructuring for getting single fields, although I personally find it easier to look at when getting multiple fields, especially when they get used multiple times in the scope the destructuring occurs.

PR #46, opened 2024-02-11
Fix minor typos in readme

sunaurus:
Just noticed a couple of minor typos


sunaurus:
Now I noticed that `run.sh` is no longer used, so I fixed the instructions for running locally as well

SleeplessOne1917 (approved):
Thanks!

PR #50, opened 2024-03-03
Lemmy errors, SSR errors, Docker dev environment

jim-taylor-business:
- use LemmyErrors from Lemmy
- reinstate SSR error handling after bug fix from Leptos
- add Docker localhost Lemmy dev environment


docker/docker-compose.yml
```diff
@@ -0,0 +1,132 @@
...
+      # port 10002, path /metrics by default
+      - "10002:10002"
+    volumes:
+      - ./lemmy.hjson:/config/config.hjson:Z
+    depends_on:
+      - postgres
+      - pictrs
+    logging: *default-logging
+
+  lemmy-ui:
```
dessalines, line 57: This shouldn't be necessary here.

It should have lemmy-ui-leptos in its place.

docker/lemmy.hjson
dessalines, line 1: You shouldn't need all this, check out [this](https://github.com/LemmyNet/lemmy-ansible/blob/main/examples/config.hjson) for a more minimal template, and you probably don't need the email block from there either.

docker/docker-compose.yml
dessalines: Or if its best to develop outside of docker (which I'm sure it is), then just leave out this block, and maybe have a comment. And it'd be good to rename this docker-compose-develop.yml or something to show that its only for development.

jim-taylor-business: i left the default file name as it makes it easier to run with docker compose, and it is pretty obvious what it's for in this context

PR #40, opened 2024-01-13
Homepage, progressive enhancement, error handling and more

jim-taylor-business:
this PR resolves #2 and resolves #8 while acknowledging that more precise issues will occur in both Home and Errors  

### features:

- updated Leptos to 0.6 and re-enabled Leptos wasm error reporting

this has resolved the performance and almost all navigation crashes so a much better experience. leptos will report useful errors when it crashes also. some side effects listed below

- full progressive enhancement from SSR-only (no-JS browser) to CSR-only (mobile/desktop)

server function interactions are only used in SSR-only mode and will be Hydrated in most browsers and function completely in CSR only app mode without server functions or any SSR. this has been enable by reestablishing isomorphic Lemmy client and cookie handling (actix-session was removed because it did not provide this; leptos-query was removed as it has been crashing server and is not necessary at this stage)

- error and validation handling with real rust error types from both Lemmy and Lemmy UI Leptos

errors are (de)serialised through the server interactions and the Lemmy client to enable this. error types can then be matched upon to display messages, decorate the page, perform actions depending on the nature of the interaction. **_i think Lemmy error types and utility functions should be directly imported from a Lemmy crate;_** at the moment they are duplicated.

- upgrade tailwind and daisy

a Leptos environment variable (LEPTOS_TAILWIND_VERSION=v3.4.1) may be necessary to use the latest tailwind

### known issues:

- SSR redirects are crashing the server

so SSR errors are handled but cannot be displayed because the querystring cannot be changed using redirects

-  browser caching depending on usage

the lemmy server i am testing against ds9.lemmy.ml does not return cache-control headers which means that in combination with my vanilla browser settings anything could be cached, including the `site` info request that determines whether or not you are logged in. for this reason there is cache busting code (a simple time signature query-param) that aims to bust that caching.

- this has all the building blocks in place for the features above, but will be missing parts and can be improved upon

i have changed a lot and i'm sure it's not perfect :) i already have improvements in mind but i need to draw a line and get this code in

PR #51, opened 2024-03-13
Add dependency on lemmy-client

### LINKED ISSUE #49: Add dependency on lemmy-client
Now that there is a rust library specifically for making requests for lemmy, it makes sense to shift to using that instead of the hand-spun one currently used in this repo.

SleeplessOne1917:
Closes #49.

This is a big one, so I'll give some more info so this is more digestible.

# Resource handling
I went back to the `leptos-query` approach of resource handling. To reiterate [what I said in a comment in the last big PR](https://github.com/LemmyNet/lemmy-ui-leptos/pull/40#discussion_r1499271213):
> As for why I'm putting up so much resistance to ditching leptos-query, it's mainly due to the fact that it solves several problems that we will run into. Caching and invalidating caches is a big plus, and I believe you are trying to accomplish something to that effect with the cache busting query param you added. Another major plus is its resource fetching deduplication. Whereas preventing resource duplication would require a lot of fiddling with contexts, the query approach makes it easy to write a server function and query in one file and use it in many different components without context management or duplication being a concern. Finally, I didn't run into the same issues with the library that you did.

In addition to above, I also found the code to be shorter and easier to read after using this approach in combination with some other things I will get to shortly.

# Cookies
I'm using `actix-session` again for this. Besides (in my opinion) being easier to work with than the existing approach, it allows us to always use `HttpOnly` in the session cookie, making it more secure as it mitigates common XSS attacks. There were 2 other things that used cookies that I removed/commented out that I must also touch on:

## Language
I don't think it makes sense to use a cookie for this. This is the sort of thing that can be derived, in order of precedence, from the user language, site language, and browser language. A cookie doesn't solve any problems here.

## Theme
This is a more interesting issue. I would normally prefer to use a cookie for the theme, but the fact that theme is also tracked as a user setting and site setting in the backend makes me wonder if we should use an approach more similar to the existing inferno UI. This concern would be easier to disregard if this was a third-party UI, but the fact that this is meant to replace the default/official UI makes me think we should honor that.

This conversation can be extended to many of the other display related settings as well, e.g. post listing types (maybe a frontend wants to have its own post listing types, or maybe a user would prefer card view on their mobile app and list view on their browser).

# Errors
Had to comment them out because they conflicted with what I was implementing. Will have to revisit them before this PR is taken out of draft.

# Unpack
This pairs quite nicely with leptos query. It allows use to use the item from a query - or signals derived from one - to work cleanly and concisely with suspenses and error boundaries.

# Misc.
- No more browser console errors (on my machine at least)!
- Deleted `Layout` and moved markup to surround `Routes` component instead since it wraps all the routes anyway
- Added `hash-files` cargo leptos option, which adds a hash to the CSS, JS, and WASM files that get served between each build


jim-taylor-business:
i know this is WIP but just to save time, i disagree with everything you've said and done in this PR for reasons i have stated in other discussions we have all had together

it's obvious we have a fundamental misunderstanding of how this stuff works. this is unfortunate to say the least, sorry, and re-writing the whole app in this manner is not something i can get behind

focusing on solving #49 only would be more useful IMO

dessalines:
What's negative about [leptos-query](https://github.com/gaucho-labs/leptos-query), it seems to have a lot of benefits.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "devDependencies": {
-    "daisyui": "^4.6.0",
-    "tailwindcss": "^3.4.1"
-  }
+    "daisyui": "^4.7.3"
+  },
+  "packageManager": "pnpm@8.15.4"
```
SleeplessOne1917, line 5: The latest version of `pnpm` adds this when I install or update deps for some reason.

src/ui/components/home/home_activity.rs
```diff
@@ -1,405 +1,52 @@
...
-                  error.set(Some(e));
-                }
-              }
-            },
-          );
-        }
-      };
-
-      window_event_listener_untyped("scroll", on_scroll);
-    }
```
SleeplessOne1917, line 226: I don't understand why we were using different fetch limits for different screen sizes. As for the infinite scroll logic, [there is a library with various leptos utility functions](https://docs.rs/leptos-use/latest/leptos_use/fn.use_infinite_scroll.html) that would greatly simplify things.

src/ui/components/common/unpack.rs
```diff
@@ -0,0 +1,13 @@
+use leptos::*;
+
+#[component]
+pub fn Unpack<T: Clone + 'static, F: Fn(T) -> Fragment + 'static>(
+  #[prop(into)] item: MaybeSignal<Option<Result<T, ServerFnError>>>,
+  children: F,
+) -> impl IntoView {
+  Signal::derive(move || match item.get() {
+    Some(Ok(item)) => Some(Ok(children(item))),
+    Some(Err(e)) => Some(Err(e.clone())),
```
SleeplessOne1917, line 10: This branch gets caught by error boundaries.

jim-taylor-business:
> What's negative about [leptos-query](https://github.com/gaucho-labs/leptos-query), it seems to have a lot of benefits.

it does work without it

jim-taylor-business:
i've said all this before but here's my disagreements with this again:

- there are no data caching solutions identified that it provides, it's just adding boilerplate and complexity and could easily be added later when caching is identified as a requirement
- the caching _**problem**_ i had to solve was a server issue between the server and the bare metal browser that forced caching
- the cookies were useful in non authenticated mode so no session has been established. IIRC the type of cookie being used could be specified in the constructor
- i hope it works well CSR only mode of operation where there is no leptos server
- i hope it works well with SSR only mode also
- not sure i can help with understanding leptos context signals they seem pretty straight forward to me

i hope things are more stable now also, these extra libraries caused show stopping issues in the past

src/ui/components/home/home_activity.rs
jim-taylor-business: more screen space, more content

SleeplessOne1917:
> there are no data caching solutions identified that it provides, it's just adding boilerplate and complexity and could easily be added later when caching is identified as a requirement

I think I'm starting to get a better understanding of the issue you have with my approach. If I understand you correctly, regardless of whether or not the caching and invalidation functionality provided by lemmy-client works/is useful, we haven't gotten to a point where those are issues yet and my usage of the library is a premature optimization. Is that an accurate interpretation?

> the caching problem i had to solve was a server issue between the server and the bare metal browser that forced caching

Looking back at the PR you introduced that change in, the problem you were solving was ds9.lemmy.ml not sending back cache control headers. This isn't a problem when testing against a locally running instance. @dessalines is the the lack of cache control headers in ds9's responses a deliberate configuration choice or accidental misconfiguration?

> the cookies were useful in non authenticated mode so no session has been established. IIRC the type of cookie being used could be specified in the constructor

Makes sense considering cookies were also being used for language and theme. I've detailed my thoughts on using cookies for those in the OP. I think the `actix-session` is easiest when dealing only with authentication, however how currently set theme should be tracked is something that needs to be ironed out and goes outside the scope of this PR (and even this project).

> i hope it works well CSR only mode of operation where there is no leptos server
If you're asking if it can be run with trunk, I haven't tested it with that yet and will need to get back to you. However, I don't know why we would want to rely on CSR only as opposed to a SSR page that gets hydrated with client side reactivity.

> i hope it works well with SSR only mode also
I've tested it out with javascript and wasm disabled on my browser and everything supported so far works (will attach screen recording when I get the chance).

> not sure i can help with understanding leptos context signals they seem pretty straight forward to me
I'm not having trouble understanding them. The use of contexts and leptos-query isn't mutually exclusive however. I've even added a context for site state that uses the advantages of leptos-query's caching/invalidation while removing the need for creating a query scope everywhere it is used.

jim-taylor-business:
> I think I'm starting to get a better understanding of the issue you have with my approach. If I understand you correctly, regardless of whether or not the caching and invalidation functionality provided by lemmy-client works/is useful, we haven't gotten to a point where those are issues yet and my usage of the library is a premature optimization. Is that an accurate interpretation?

yes, it's a solution looking for a problem _**at the moment**_. it has caused all sorts of problems for me in the past. i'm guessing it will not solve the caching problem. i haven't seen it progressively enhance to CSR only mode. so at the moment it obfuscates server communication and the vanilla leptos version works fine without it.

> This isn't a problem when testing against a locally running instance. @dessalines is the the lack of cache control headers in ds9's responses a deliberate configuration choice or accidental misconfiguration?

i have replicated this locally. i totally agree that if we are going to use the `site` `get` request to tell us important information like whether we are authenticated or not then the server should add a `pragma no-cache header`. will submit PR with some e2e tests to test for it locally today. it's very timing and maybe browser sensitive but i'm sure the cache header would fix it. might submit a PR to the server project to get some discussion going over there.

> Makes sense considering cookies were also being used for language and theme. I've detailed my thoughts on using cookies for those in the OP. I think the `actix-session` is easiest when dealing only with authentication, however how currently set theme should be tracked is something that needs to be ironed out and goes outside the scope of this PR (and even this project).

i agree the final solution has not been defined yet and will involve the lemmy server significantly so what is there is just useful for testing and i think, for now, is a nice feature. i also don't think we need to copy what's currently happening in Lemmy UI if it is small useful addition to a feature.

> If you're asking if it can be run with trunk, I haven't tested it with that yet and will need to get back to you. However, I don't know why we would want to rely on CSR only as opposed to a SSR page that gets hydrated with client side reactivity.

CSR mode is naturally used when the same code is used for a Android/iOS/Windows/MacOS app and it is currently a declared objective of the project on the homepage so that's why i made everything work with that mode. it's a feature of Leptos generally and i think the whole isomorphic rust thing is cool.

> I'm not having trouble understanding them. The use of contexts and leptos-query isn't mutually exclusive however. I've even added a context for site state that uses the advantages of leptos-query's caching/invalidation while removing the need for creating a query scope everywhere it is used.

cool. i seem to remember i put the site state in the context as well and had some performance issues with it as it can be pretty large and is copied around a lot which is why i went for a simple boolean flag to indicate logged in status. that was a while ago though so Leptos might have optimized that better now

dessalines:
> Looking back at the PR you introduced that change in, the problem you were solving was ds9.lemmy.ml not sending back cache control headers. This isn't a problem when testing against a locally running instance. @dessalines is the the lack of cache control headers in ds9's responses a deliberate configuration choice or accidental misconfiguration?

Those are set in the back-end, and don't look like they're being overridden in nginx, but I do see that ds9.lemmy.ml is running a test back-end version. voyager.lemmy.ml is on `0.19.3`.

jim-taylor-business:
> Those are set in the back-end, and don't look like they're being overridden in nginx, but I do see that ds9.lemmy.ml is running a test back-end version. voyager.lemmy.ml is on `0.19.3`.

the reason the `site` cache bug is not happening in Lemmy UI is because it does not refresh the `site` object when you logout, only when you log in.

the browser cache defaults to "private" when you're logged in. i guess because there is a auth token header set at that time.

when you are logged out the cache default is 60s which is why the bug occurs when you immediately log back in. 

the work around is not to refresh the `site` information on logout, just clear the `my_user` information which i guess is roughly what happens in Lemmy UI

SleeplessOne1917:
@jim-taylor-business now that I better understand your gripe with leptos-query, I think we're on the same page. At this point, I could rewrite what I currently have to use vanilla resources, though I want to hold off pulling the trigger unless smoke testing of the app reveals that it's causing problems.

Regarding the caching issues you mentioned: is there a way to consistently reproduce the bug? It could very well be a browser issue as you said. I've tested my changes on Firefox, Chromium, and Librewolf and haven't run into the issue.

About CSR: I haven't tested that yet and definitely need to make sure my changes work there (which I doubt they do at the time of making this comment). Do you use trunk to test CSR? Or is there some feature of cargo leptos that lets me use CSR instead of SSR with hydrate?

Finally, I will re-add the cookie stuff for the languages and themes, although I would like to keep the actix session approach for handling the auth cookie.

jim-taylor-business:
> @jim-taylor-business now that I better understand your gripe with leptos-query, I think we're on the same page. At this point, I could rewrite what I currently have to use vanilla resources, though I want to hold off pulling the trigger unless smoke testing of the app reveals that it's causing problems.
> 
> Finally, I will re-add the cookie stuff for the languages and themes, although I would like to keep the actix session approach for handling the auth cookie.

so am i right in saying that the only reason these two libraries are being used is that you are convinced they will be useful in the future?

> Regarding the caching issues you mentioned: is there a way to consistently reproduce the bug? It could very well be a browser issue as you said. I've tested my changes on Firefox, Chromium, and Librewolf and haven't run into the issue.

the cache bug details are in my post just above and are reproduced in the end to end tests. maybe something you've already done has fixed it.

> About CSR: I haven't tested that yet and definitely need to make sure my changes work there (which I doubt they do at the time of making this comment). Do you use trunk to test CSR? Or is there some feature of cargo leptos that lets me use CSR instead of SSR with hydrate?

yeah you can use `trunk`

SleeplessOne1917:
I removed leptos-query from the dependencies for now. I thought I needed it to manually refetch the resources, but it turns out resources already have that functionality. I'll see what I can do about removing actix-session. I will be shifting my priority to the JS UI for the next few days to get it ready for 0.19.4.

jim-taylor-business:
cool, good luck with Lemmy UI

i'm just interested to know what your idea is for the design in future and how these crates play a role, because i do recognise that they do things that could be useful and i just can't think of the features or design where that happens

dessalines:
No probs, and no rush on this of course, but it still would be nice to have this dependent on your lemmy-client-rs, I think we're all in agreement there.

SleeplessOne1917:
@jim-taylor-business How did you run the project with CSR before? Trunk's docs mention that it needs to have an `index.html`, but there was no such file before I started this PR.

dessalines:
ping me when you want me to take another look.

jim-taylor-business:
> @jim-taylor-business How did you run the project with CSR before? Trunk's docs mention that it needs to have an `index.html`, but there was no such file before I started this PR.

I actually tested it as part of a Tauri app and generated some HTML and CSS in that project. I found that leptos handles the transition to CSR only very gracefully as long as you don't force the use of the leptos server functions

jim-taylor-business:
@SleeplessOne1917 i can't see any handling of the errors returned by the lemmy_client or any user feedback

also the login goes into some kind of infinite loop on a bad login.

i guess this is still WIP as i notice the pipeline tests are not passing?

jim-taylor-business:
> @dessalines trying to make this work in CSR while using the server actions/functions as a single source of truth for making API requests lead me down a rabbit hole. You can see an experiment I made investigating this here: https://github.com/SleeplessOne1917/trunk-cargo-leptos-example. If you're fine needing to go through some inconvenient steps to test this locally, I think this is good to take out of draft.

@SleeplessOne1917 in CSR mode the Leptos SSR server does not exist. Trunk is only used to test this because it is a simple web server. 

when you are running the code on the desktop or mobile app Trunk is not guaranteed to exist either. the HTML and CSS will be hard coded and everything will be rendered in the browser and the only server to be communicated with will be the Lemmy instance.

SleeplessOne1917:
> I actually tested it as part of a Tauri app and generated some HTML and CSS in that project.

That explains why I couldn't get it working with Trunk out of the gate. 

> is there any chance you could make a new summary of what's changed just to provide an overview? I'm guessing the summary at the beginning of this PR is out of date.

I'll look over the changes again and make a review either later tonight or tomorrow. The scope of this PR definitely crept while I was working on it.

> @SleeplessOne1917 in CSR mode the Leptos SSR server does not exist. Trunk is only used to test this because it is a simple web server.
> when you are running the code on the desktop or mobile app Trunk is not guaranteed to exist either. the HTML and CSS will be hard coded and everything will be rendered in the browser and the only server to be communicated with will be the Lemmy instance.
> that's why the client moves to using the direct communication to the lemmy instance, it doesn't need the server function as a proxy. you can easily share common functions (sources of truth) between the two.

I admittedly went down a rabbit hole trying to figure out how to reuse server actions between all of our use cases to no avail. I'll shift closer to what you were originally doing for CSR, although I'm going to try using client side actions before trying regular on submit handlers.

Also, are we set on Tauri as the desktop app framework? I know in the README it says we've yet to agree on a framework for that, but it's the only framework I know of that lets us use our webapp code for a desktop app. The closest thing to alternative that meets that criterion and comes to my mind is if we wait for if/when Leptos supports being use with GTK or another UI library, but the closest thing that exists to that currently is a tiny demo example app. If we are going with Tauri, we may have a better way of storing sessions/selected theme than using cookies since we'll have access to more than just the browser APIs.

SleeplessOne1917:
Per @jim-taylor-business's request, here is a list of changes.

# Additions
- `lemmy-client` for requests
- Theme enum for themes
- `Unwrap` component to make it easier to use resources with `Suspense`s and `ErrorBoundary`s
- A cookie middleware for handling session cookies
- Contexts for current site state and theme
- `derive_query_signal` fn to make it easier to derive signals from resources that use server fns
- `derive_user_is_logged_in` to easily get whether to use is logged in from the site state.

# Removals
- Client side cookie handling (will go into this more in a minute)
- Error types specific to our app (I think matching off of the `LemmyErrorType` returned in failed requests works for now, and any errors specific to this frontend app alone can be addressed once we reach a point we need to do it)
- Infinite scroll logic
- CSR event handling/resource fetching

After working and thinking on this awhile, I'm starting to question if we should be worrying about CSR at this point of development. We know for certain that this is going to be used in the browser, so handling SSR and hydrated cases is a given. However, how strongly decided are we on making this into a desktop app as well? I ask because having access to more than just the browser APIs on the frontend likely warrants doing some things differently. For example, it could be the case that something like [tauri-plugin-stronghold](https://crates.io/crates/tauri-plugin-stronghold) is a more appropriate way of storing login sessions than cookies. I consulted both the Leptos and Tauri discords on the subject and there is no preferred/idiomatic way of doing this. It doesn't help that tauri + leptos examples and templates are few and far between, and none I found demonstrated managing sessions.

Since this is already a big initiative with it being a re-implementation of a large existing frontend we want to reach feature parity with, and since how to handle a leptos app with tauri is currently sparse on examples we can reference and completely undocumented, I think it would be wiser to focus on making this a working webapp for use in browsers (so SSR and hydrate) first and refactor if we ever end up using this for a desktop app as well.

dessalines:
I've had no time to work on this tho, so I really don't get a say.

IMO we should prioritize this as an SSR app. This is still a web framework, not "native" at all. If someone wants to create a native app they should use a desktop UI framework like GTK or QT or one of the rust cross-platform-ish ones, or else just create a thin web wrapper that calls the SSR. A javascript SPA (which is what the CSR would be), doesn't really make sense IMO.

SleeplessOne1917:
Another thought occurred to me that further makes me think keeping this an SSR/hydrated app that runs in the browser is the right idea. One of the things that sets Leptos apart from many web UI frameworks that currently exists is its support for islands architecture. Since this would involve components that are only ever statically rendered on the server, and since components in islands architecture can run serverside code, it's not compatible with CSR. Using islands architecture will allow us to substantially reduce the size of WASM bundles sent to the browser, so I believe it is a good feature to pursue.

jim-taylor-business:
> I admittedly went down a rabbit hole trying to figure out how to reuse server actions between all of our use cases to no avail. I'll shift closer to what you were originally doing for CSR, although I'm going to try using client side actions before trying regular on submit handlers.

Yeah that sounds great. I was recently starting to try to think of a reusable component or something that encapsulates the SSR > Hydrate > CSR transition, and maybe even error handling, more cleanly without all that seemingly repititive boilerplate.

> Also, are we set on Tauri as the desktop app framework?

No i just did that for fun to prove it could be done.

> IMO we should prioritize this as an SSR app. This is still a web framework, not "native" at all. If someone wants to create a native app they should use a desktop UI framework like GTK or QT or one of the rust cross-platform-ish ones, or else just create a thin web wrapper that calls the SSR. A javascript SPA (which is what the CSR would be), doesn't really make sense IMO.

There are many successful performant apps these days using web-like frameworks and we have the advantage of using Rust in WASM and maybe even WebGPU in future. And with frameworks like Tauri we kind of get it "for free".

> Using islands architecture will allow us to substantially reduce the size of WASM bundles sent to the browser, so I believe it is a good feature to pursue.

Yeah that seems possible but I guess at the moment I'm a fan of keeping things simple, but worth looking at.

SleeplessOne1917:
Does anyone think this PR is in a good spot? Besides addressing responses to issues with testing and code review, I'm not sure what other changes I need to make.

> > Using islands architecture will allow us to substantially reduce the size of WASM bundles sent to the browser, so I believe it is a good feature to pursue.

> Yeah that seems possible but I guess at the moment I'm a fan of keeping things simple, but worth looking at.

Agreed, although I consider CSR/desktop app support the same as islands in this regard.

jim-taylor-business:
> @dessalines @jim-taylor-business Does anyone think this PR is in a good spot? Besides addressing responses to issues with testing and code review, I'm not sure what other changes I need to make.

in my experience when you rewrite something it's generally accepted that the code will be improved, and maybe even extended with new functionality. so removing one of Leptos' core abilities because there was a particular design choice made at the start of the PR is bad. error handling and some of the dependent signalling seems to have taken a hit too but i realize this is still a WIP. here's the code working unchanged in a demo Tauri project that compiled to an Android APK automatically to prove the power of Leptos and Tauri.

https://github.com/LemmyNet/lemmy-ui-leptos/assets/7439891/a075f268-50f6-48ed-b208-093de579fcb5

the general refactoring of components and the addition of the lemmy client and Theme enum are nice and i guess the other tools are useful.

dessalines:
It looks good to me, and especially in these early stages, we should be merging PRs rapidly and getting the infrastructure in place, so that we can start building out the components quickly after.

We can go to a stable toolchain anytime (provided we aren't using anything in leptos that requires it), that's not a blocker for this.

Try to get the linting passing, then we can merge.

dessalines (approved):

dessalines:
lemmy-ui is an isomorphic app, so it does proxy the initial page load through a node server, but the node server itself doesn't have to set up endpoints in the same way leptos does. 

I agree it really doesn't make sense to use [leptos server functions](https://book.leptos.dev/server/25_server_functions.html), since we aren't doing anything other than making http calls. Those should only be if you're building a full app, need to make DB calls, etc.

@SleeplessOne1917 what's the reason again that this layer is required?

SleeplessOne1917:
The reason it heavily relies on server functions is to take advantage of the progressive enhancement/graceful degradation they allow. I could very well set up the app to call the backend server directly from the browser using the lemmy client directly via event handlers (e.g. `on:click`, `on:submit`, etc.). However, if I did that, the app would only be usable if JS and WASM are enabled. Using server functions guarantees the logic will work both with and without JS enabled without needing to duplicate code.

Also, there is one thing that is only accessible on the server side: the session cookie. [I made the cookie HttpOnly to help mitigate potential XSS attacks](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#block_access_to_your_cookies). This makes it so the cookie cannot be read from scripts in the browser, making server-side logic necessary.

SleeplessOne1917:
I just had another thought regarding native apps: would any of the platform native APIs (e.g. window, filesystem, clipboard, etc.) give us enough benefits over what we can already do in the browser that make a native app worth the effort. If we don't end up needing any of that, I think a PWA will be sufficient for our purposes.

jim-taylor-business:
> I could very well set up the app to call the backend server directly from the browser using the lemmy client directly via event handlers (e.g. `on:click`, `on:submit`, etc.). However, if I did that, the app would only be usable if JS and WASM are enabled. Using server functions guarantees the logic will work both with and without JS enabled without needing to duplicate code.

not true. if you look at the old demo i made you'll notice that the code worked in SSR (no JS/wasm) and CSR (no serverfns) unchanged, @SleeplessOne1917 @dessalines. the logic of the interactions was shared in the same rust functions that run on both server and browser, no duplication.

there was a tiny bit of extra boilerplate but that could probably be refactored out using a custom form component.

it seems like you're committed to this new design so attempting to rewrite this would not be productive

i admire that you're encouraging new contributions @dessalines but i am concerned about the value of them when they're based on incorrect assumptions

jim-taylor-business:
> I just had another thought regarding native apps: would any of the platform native APIs (e.g. window, filesystem, clipboard, etc.) give us enough benefits over what we can already do in the browser that make a native app worth the effort. If we don't end up needing any of that, I think a PWA will be sufficient for our purposes.

the real value of CSR based native apps is not proxying everything through the leptos server, in fact not needing a leptos server at all, just a lemmy server. like normal apps do.

dessalines:
If I were going to make a native desktop app for lemmy, I wouldn't use leptos or tauri, but probably [iced](https://iced.rs/), or another real desktop toolkit.

The thin layer that leptos has us add is pretty minor IMO. The leptos page on [progressive enhancement](https://book.leptos.dev/progressive_enhancement/index.html) convinced me that there's no reason to support a CSR / SPA mode, and adding their layer gives a lot more benefits.

I agree that most normal apps shouldn't need this (or any layer) really, all computer programs *should* be local, but the browser makes us throw these things all out the window with its multi-step process to load a program then make calls, which is why so many web frameworks are SSR / isomorphic now.

Anyway I do think we should go forward with leptos' recommendations for SSR apps, and use their `<ActionForm`

SleeplessOne1917:
Another promising option for a native UI framework is [GPUI, the framework made by the devs of the Zed editor](https://www.gpui.rs/). It's still early in development, but since our current priority is making a web app to run in the browser, I expect the framework to be more fleshed out by the time we expand to a native desktop app.

jim-taylor-business:
> The thin layer that leptos has us add is pretty minor IMO. The leptos page on [progressive enhancement](https://book.leptos.dev/progressive_enhancement/index.html) convinced me that there's no reason to support a CSR / SPA mode, and adding their layer gives a lot more benefits.

progressive enhancement and [ActionForm](https://book.leptos.dev/progressive_enhancement/action_form.html)s do not stop upgrading the interactivity with server functions:

> > Because the `<ActionForm/>` is just a `<form>`, it fires a submit event. You can use either HTML validation, or your own client-side validation logic in an on:submit 

...and re-use the code to make the lemmy server interaction (because it's already isomorphic!) and take the load off the web server by not proxying through it unnecessarily

> If I were going to make a native desktop app for lemmy, I wouldn't use leptos or tauri, but probably [iced](https://iced.rs/), or another real desktop toolkit.

you could re-write the whole UI again or get it almost for free with Tauri

you seem to be implying that you are forced to use server functions because this is a web project but this doesn't happen in Lemmy UI

jim-taylor-business:
well, @dessalines, it seems like you're following Sleepless' plan for everything right now so we'll see where we end up

PR #54, opened 2024-04-03
end 2 end woodpecker step

jim-taylor-business:
- used playwright image to speed up install
- aligned client and test servers on the 0.19.4 beta
- align client and tests on stable rust except for code format
- update tooling versions (fixes ssr leptos bug, etc)

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -94,7 +94,7 @@ steps:
       - pnpm install
 
       - rustup target add wasm32-unknown-unknown
-      - LEMMY_UI_LEPTOS_LEMMY_HOST=lemmy LEMMY_UI_LEPTOS_LEMMY_HTTPS=false LEPTOS_TAILWIND_VERSION=v3.4.1 cargo leptos watch &
+      - LEMMY_UI_LEPTOS_LEMMY_HOST=lemmy:8536 LEMMY_UI_LEPTOS_LEMMY_HTTPS=false LEPTOS_TAILWIND_VERSION=v3.4.1 cargo leptos watch &
 
       - cd end2end
       - pnpm install
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -158,7 +158,7 @@ steps:
 
       - rustup target add wasm32-unknown-unknown
       - LEMMY_UI_LEPTOS_LEMMY_HOST=lemmy:8536 LEMMY_UI_LEPTOS_LEMMY_HTTPS=false LEPTOS_TAILWIND_VERSION=v3.4.1 cargo leptos watch &
-      # - sleep 30
+      - sleep 30
 
       - cd end2end
       - pnpm install
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -1,5 +1,5 @@
 variables:
-  - &rust_image "rust:latest"
+  - &rust_image "rust:nightly"
   - &install_binstall "wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz && tar -xvf cargo-binstall-x86_64-unknown-linux-musl.tgz && cp cargo-binstall /usr/local/cargo/bin"
 
 services:
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -54,13 +54,13 @@ steps:
       - event: pull_request
 
   cargo_fmt:
-    image: *rust_image
+    image: rustlang/rust:nightly
     environment:
       # store cargo data in repo folder so that it gets cached between steps
       CARGO_HOME: .cargo
     commands:
       - rustup component add rustfmt
-      - cargo +nightly fmt -- --check
+      - cargo fmt -- --check
     when:
       - event: pull_request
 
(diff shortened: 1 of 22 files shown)


jim-taylor-business:
as you can see @dessalines i have made a few attempts to run an e2e test by trying to get a woodpecker instance of lemmy running against a database. i could mock all the lemmy end points in the tests but that wouldn't be an e2e test to me, more of an integration test of some sort.

i tried running docker in docker in woodpecker but that requires a privileged container which is not great for security of the woodpecker server.

i also tried to run the lemmy:dev image as a woodpecker service but it won't work without a config file because of this line:

https://github.com/LemmyNet/lemmy/blob/087684658a5d2008dff3df54d9b6e0ebde1ce1bc/crates/utils/src/settings/mod.rs#L16

i have changed the `expect` to an `unwrap_or_default` and set the `LEMMY_DATABASE_URL` and the lemmy instance starts, connects to the DB and works in my tests locally. do you think this is a reasonable change to make to lemmy to allow it to be used without a config file in woodpecker @dessalines?

jim-taylor-business:
see https://github.com/LemmyNet/lemmy/pull/4594 for suggested implementation

Nothing4You:
this can also be done without changes to the lemmy container:

https://github.com/Nothing4You/lemmy-ui-leptos/commit/91badd20ad757eee7c20bdc727248e134f6ffce5
https://wpci.lem.rocks/repos/2/pipeline/14/5

the sleeps could be replaced by checking if the service is available yet, this was just a quick n dirty poc.

jim-taylor-business:
ooh clever! i totally forgot about commands in the container

Nothing4You:
the woodpecker docs are unfortunately a bit limited, i ended up resorting to the json scheme definition: https://github.com/woodpecker-ci/woodpecker/blob/7b731d6b9ee146f13068e2836b0c80b2faef4542/pipeline/frontend/yaml/linter/schema/schema.json#L802

jim-taylor-business:
also had to work around cargo-leptos also apparently always crashing at the end of the test run and exiting with a fail

.woodpecker.yml
```diff
@@ -1,7 +1,43 @@
 variables:
-  - &rust_image "rustlang/rust:nightly"
+  - &rust_image "rust:latest"
```
dessalines, line 2: You can just use `rust` then, since its just the stable tag.

Is there any reason you can't run nightly tho?

.woodpecker.yml
```diff
@@ -1,7 +1,43 @@
...
   - &install_binstall "wget https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz && tar -xvf cargo-binstall-x86_64-unknown-linux-musl.tgz && cp cargo-binstall /usr/local/cargo/bin"
 
+services:
+  lemmy:
+    name: lemmy
+    image: dessalines/lemmy:0.19.4-beta.2
+    commands:
+      - |-
+        printf '{
```
dessalines, line 11: Add this as a file somewhere, probably in your new test folder.

.woodpecker.yml
```diff
@@ -23,21 +59,26 @@ steps:
       # store cargo data in repo folder so that it gets cached between steps
       CARGO_HOME: .cargo
     commands:
+      - rustup component add rustfmt
```
dessalines, line 62: We do `cargo +nightly fmt` on all lemmy repos, so you do need to use the `rust:nightly` image for this.

.woodpecker.yml
jim-taylor-business: just matching how it is currently compiled in dev

is there a lot of benefits with nightly?

.woodpecker.yml
jim-taylor-business: the woodpecker pipeline would need to be privileged to map a file into the service which is why this workaround is needed if i understand correctly

.woodpecker.yml
jim-taylor-business: ok i will switch back to the nightly image

.woodpecker.yml
dessalines: See L27, you should be able to copy or symlink the file into the correct location, and it should work. All the files should be accessible to the images.

jim-taylor-business: i think i tried this in this commit

https://github.com/LemmyNet/lemmy-ui-leptos/pull/54/commits/d70a9b39d5ab9a0b89d7f406952c8685d099d3aa

and the pipeline refused to run because of permissions but i will give it another go

jim-taylor-business: here's the documentation that i think applies if i'm getting this right:

https://woodpecker-ci.org/docs/usage/volumes

jim-taylor-business: oh sorry i think i understand now with pipelines the src is in the container file system! will see if this works with services

jim-taylor-business: ok this is done, thx @dessalines 

just waiting for the pipeline's network issues to recover to get a good build

dessalines (approved):
Sweet, thx!

PR #55, opened 2024-05-09
Style and markup tweaks, break up post listing into smaller components

SleeplessOne1917:
# Changes
- Break up `PostListing`, specifically extracting content actions and vote buttons into components that can be reused between posts and comments, similar to what we do in the inferno UI.
- Use unordered list for `PostListings` instead of table, as the data is not tabular.
- Fix theming bug by setting `data-theme` on root `html` element instead wrapping `div` (see 4th picture in each section below)
- Fix `get_post` server fn
- Use different upvote and downvote icons. Experimenting with this a bit. [Here are some other arows from the icon set we're using if you want to see what else is out there.](https://phosphoricons.com/?q="arrow")

To show some of the visual changes:

## Old
old_home_big_screen
old_home_small
old_home_smaller
old_dev

## New
new_home_big
new_home_small
new_home_smaller
new_dev_tools

## Rough Edges
- The header and sidebars are currently not responsive. I have ideas for them beyond trying to replicate what the inferno UI does, so I'm deferring them for another PR.
- The post listing looks a bit odd in `PostActivity`, so I'll experiment with making that look better for post pages.
- The content action components only have a few actions and need to be fleshed out in both functionality and style

I will address these issues in future PRs.


src/ui/components/common/content_actions/post_content_actions.rs
```diff
@@ -0,0 +1,80 @@
...
+    .save_post(LemmyRequest {
+      body: SavePostBody { post_id: id, save },
+      jwt,
+    })
+    .await
+    .map_err(|e| ServerFnError::ServerError(e.to_string()))
+}
+
+#[server(prefix = "/serverfn")]
+async fn report_post(post_id: PostId, reason: String) -> Result<PostReportResponse, ServerFnError> {
```
dessalines, line 29: Just a note that it'd probably be easier to just pass the form objects in. That's what we do in jerboa for the most part.

These should also probably live in different files. IE separate your UI components from your API actions, IE


- a file for viewmodels / API actions / data persisted across different screens
- A file for the "screen" or main route with all the query-related stuff
- A file(s) for UI components.

src/ui/components/common/vote_buttons/comment_vote_buttons.rs
```diff
@@ -0,0 +1,39 @@
...
+};
+use leptos::*;
+
+#[server(prefix = "/serverfn")]
+async fn vote_comment(id: CommentId, score: i16) -> Result<CommentResponse, ServerFnError> {
+  use crate::{constants::AUTH_COOKIE, utils::get_client_and_session};
+
+  let (client, session) = get_client_and_session().await?;
+
+  let jwt = session.get::<String>(AUTH_COOKIE)?;
```
dessalines, line 17: Might be helpful to add another impl for session, with a `session.get_jwt()` function.

dessalines (approved):
Nice, feel free to merge whenever you see fit.

src/ui/components/common/content_actions/post_content_actions.rs
SleeplessOne1917: Agreed. I have similar, yet slightly different, ideas for organizing things. I'll open an issue detailing my ideas when I get the chance later today.

dessalines: Cool

PR #57, opened 2024-05-09
Dockerize end to end tests so they can be run locally

SleeplessOne1917:
I made this a draft for a few reasons:

- So far the dockerfile only builds an image that needs to be run separately
- There needs to be a backend server in order for the login tests to work
- Can we reuse the dockerfile for this test for woodpecker?
- No clue what to do with existing lemmy.hjson in end2end folder
- Should we write reports of runs to a volume to review?

My docker fu is nothing to write home about, so I would appreciate any help given.


end2end.dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/playwright:v1.44.0-jammy
```
SleeplessOne1917, line 1: Need to make this match the version used in `end2end/package.json` instead of using the latest tag.

dessalines:
I should be able to get to this tmrw, but there's probably a few examples of dockerized playwright tests you can find online. 

Having an actual running lemmy server makes it more complicated, but we do have an example in lemmy's `.woodpecker.yml` of how to run and test end to end test lemmy within docker.

SleeplessOne1917:
@dessalines update on this:

I've got the CI/CD pipeline and local end to end test script working. With that done, I'll add some more end to end tests and see if I can get the tests running in parallel.

I've learned a lot working on this PR, and I can use that knowledge to improve some of our other woodpecker pipelines in the future.

SleeplessOne1917:
This is ready to go.

scripts/run_end2end_tests.sh
```diff
@@ -0,0 +1,25 @@
+# This script assumes docker can be run as a non-root user. See the following docs if the script fails because you need to use sudo:
+# https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user
+
+CONTANIER_NAME='end2end'
```
dessalines, line 4: `CONTAINER`

dessalines (commented):
Damn, nice job. Feel free to merge whenever you see fit.

dessalines (approved):

scripts/run_end2end_tests.sh
dessalines: Oops, my approval merged this accidentally.

PR #84, opened 2024-06-11
Spruce up the sidebar

SleeplessOne1917:
Made some visual changes to the sidebar:

Updated sidebar

Still more to be done, especially for mobile. I'll take another look at this after doing more thorough work on the mobile burger menu, filter bars, and probably post listings too.


src/utils/format_number_si.rs
```diff
@@ -0,0 +1,206 @@
...
+            1
+          },
+          number_as_float
+        );
+      }
+    }
+
+    panic!("Number {number} is larger than 1 trillion! Where are you getting numbers this big?");
+  }
+}
```
SleeplessOne1917, line 26: While this can be done with the browser's internationalization API via `js_sys`, that won't work on the server. I also noticed that the old UI was only using the `en-US` locale. If we plan to only handle that locale for these numbers, this algorithm should be fine.

src/ui/components/home/trending.rs
SleeplessOne1917, line 1: Removed this like you suggested for the old UI.

src/ui/components/common/icon.rs
```diff
@@ -34,19 +35,23 @@ pub enum IconType {
...
 
 impl IntoAttribute for IconSize {
   fn into_attribute(self) -> Attribute {
     match self {
-      Self::Large => "3rem",
+      Self::ExtraLarge => "3rem",
+      Self::Large => "2.25rem",
       Self::Small => "0.75rem",
       _ => "1.5rem",
     }
```
SleeplessOne1917, line 57: If I find myself needing another size, I'll just stick to tailwind. Maybe make some utility classes in the config.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ leptos_meta = { version = "0.6" }
 leptos_router = { version = "0.6" }
 leptos_i18n = "0"
 cfg-if = "1"
-lemmy-client = { version = "0.19.4-rc.7", features = ["leptos"] }
+lemmy-client = { version = "1.0.0", features = ["leptos"] }
```
SleeplessOne1917, line 15: [See this relevant issue if you're confused about the version change.](https://github.com/LemmyNet/lemmy-client-rs/issues/11)

src/ui/components/home/site_summary.rs
```diff
@@ -1,14 +1,53 @@
...
+  view! {
+    <tr class="*:p-2.5 [&:not(:first-child)]:border-t-2 [&:not(:first-child)]:border-accent">
+      <td class="text-xs font-semibold">{text}</td>
+      <td class="text-center font-bold">{format_number_si(count)}</td>
+    </tr>
+  }
+}
+
+#[component]
+fn AdminCard(
```
dessalines, line 23: Up to you of course, but I'd recommend just sticking with a generic `UserListing` component that can be used nearly everywhere. 

Those are nice because its uniform, you only have one to maintain, can create consistent coloring and linking, etc.

src/ui/components/home/site_summary.rs
```diff
@@ -32,52 +71,81 @@ pub fn SiteSummary() -> impl IntoView {
...
-        </Transition>
-        <h3 class="card-title">Admins</h3>
-
-        <p>
-          <Transition>
-            <Unpack item=admins let:admins>
-              <For each=move || admins.clone() key=|c| c.0 let:admin>
-                <CountsBadge>{admin.1}</CountsBadge>
-              </For>
+          <section aria-labelledby="instance-stats-heading" class="my-4">
```
dessalines, line 88: This table does look really good btw.

src/utils/format_number_si.rs
dessalines: There's a library for this also: https://docs.rs/si_format/latest/si_format/

IMO we should just use SI formatting for any potentially large numbers, and not worry about any internationalization on them.

src/utils/format_number_si.rs
```diff
@@ -0,0 +1,206 @@
...
+            0
+          } else {
+            1
+          },
+          number_as_float
+        );
+      }
+    }
+
+    panic!("Number {number} is larger than 1 trillion! Where are you getting numbers this big?");
```
dessalines, line 24: Pry won't need this once you use the library.

src/utils/format_number_si.rs
```diff
@@ -0,0 +1,206 @@
...
+  }
+}
+
+#[cfg(test)]
+mod test {
```
dessalines, line 29: Pry aren't going to need these tests when you use the library.

dessalines (approved):
Looks great so far. Feel free to merge when tests are passing.

src/ui/components/home/home_page.rs
```diff
@@ -32,8 +32,8 @@ pub fn HomePage() -> impl IntoView {
   let posts = derive_query_signal(posts_resource, |r| r.posts.clone());
 
   view! {
-    <div class="md:container md:grid md:grid-cols-5 xl:grid-cols-4 md:grid-rows-1 mx-auto mt-4 mb-1 md:gap-20 h-fit md:h-full">
-      <main class="md:col-span-3 flex flex-col mx-2.5 md:mx-0 h-fit md:h-full">
+    <div class="flex lg:container mx-auto mt-4 mb-1 lg:gap-12 h-fit lg:h-full">
+      <main class="basis-full lg:basis-[65%] xl:basis-3/4 flex flex-col mx-2.5 lg:mx-0 h-fit lg:h-full">
```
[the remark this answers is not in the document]
SleeplessOne1917: I tried the tailwind defaults and they all seemed a little off.

src/ui/components/home/site_summary.rs
SleeplessOne1917: I'm thinkg of using something like the `PersonListing` from the old UI for showing the user's info on posts, comments, and DMs. This and (possibly, depending on how it looks) the person info on the profile page I'm thinking of making different from the more generic listing.

dessalines (approved):

PR #107, opened 2024-07-28
Some post content actions, creator listings, and community listings

SleeplessOne1917:
I was going to make this everything related to posts, including all of the content actions and the post page (sans comments), however this set of changes has already gotten pretty big. Notable changes:

- Made creator and community listings for when user posts to community:
creator and post listings
The Lemmy icon is the default PFP for users when none is set. The apartment icon is the default icon for communities for when none is set.

- Added fedi link. Since I'm using a phosphor icon, which I would have a difficult time making multi-colored like the one in the inferno UI, I gave it a rainbow cycle effect on hover to still give it a bit of life:

https://github.com/user-attachments/assets/c8b42225-da58-4c4b-80a6-cea908a215c6

- Made content actions dropdown actions look a bit better:
content actions dropdown

- Added report content modal
report content

I made some smaller tweaks that I will point out in comments.


Cargo.toml
```diff
@@ -45,16 +46,16 @@ hydrate = [
   "leptos_meta/hydrate",
   "leptos_router/hydrate",
   "leptos_i18n/hydrate",
-  "dep:wasm-bindgen",
```
SleeplessOne1917, line 48: `js-sys` re-exports `wasm-bindgen`, so we don't need the extra dependancy.

src/ui/components/common/content_actions/post_content_actions.rs
SleeplessOne1917, line 1: It turns out context makes it easier to handle different content actions, making the content specific actions components unnecessary.

src/ui/components/common/vote_buttons/post_vote_buttons.rs
SleeplessOne1917, line 1: Similar story here as with content actions.

src/ui/components/layouts/base_layout.rs
```diff
@@ -19,30 +19,31 @@ pub fn BaseLayout() -> impl IntoView {
...
-            <Html attr:data-theme=theme class="h-full max-h-screen overflow-y-hidden"/>
-            <TopNav/>
+            <Html attr:data-theme=theme class="h-full max-h-screen overflow-y-hidden" />
+            <TopNav />
             <div class="flex gap-x-4 h-full overflow-y-auto md:overflow-y-hidden grow">
-              <aside class="w-fit px-3.5 whitespace-nowrap pb-5 pt-3 border-e border-neutral hidden md:flex flex-col bg-base-300 overflow-y-auto h-full">
-
-                <SideNav/>
-              </aside>
+              <nav
```
SleeplessOne1917, line 29: `aside` tags are meant to enhance another attribute while being related to it, but not neccessary to understand it. Since this is a site-wide nav, the `nav` tag is more appropriate.

src/ui/components/layouts/base_layout/side_nav.rs
```diff
@@ -11,22 +11,22 @@ pub fn SideNav() -> impl IntoView {
   let i18n = use_i18n();
 
   view! {
-    <nav aria-label="Pages nav">
+    <section aria-label="Main navigation links">
```
SleeplessOne1917, line 14: `section`s seem a better semantic choice for groups of links within a nav than several separate navs.

src/utils/apub_name.rs
```diff
@@ -0,0 +1,96 @@
...
+  create_apub_name::<'!'>(actor_id)
+}
+
+fn format_apub_name<const PREFIX: char>(instance: &str, name: &str) -> String {
+  format!("{PREFIX}{name}@{instance}")
+}
+
+fn create_apub_name<const PREFIX: char>(actor_id: &str) -> String {
+  const REGEX_STR: &str =
+    r"https?:\/\/((?:(?:[^.]\.)*[^.]+\.[^:]+)|localhost)(?::\d{1,5})?\/[uc]\/([^?#]+)";
```
SleeplessOne1917, line 18: Not a perfect regex for URL parsing, but I anticipate actor IDs to consistently be the same format, allowing us to take some shortcuts. Let me know if my assumption is wrong.

src/utils/apub_name.rs
```diff
@@ -0,0 +1,96 @@
...
+        use regex::Regex;
+        static INSTANCE_REGEX: LazyLock<Regex> = LazyLock::new(|| Regex::new(REGEX_STR)
+          .expect("Could not create server side regex.")
+        );
+
+        INSTANCE_REGEX
+          .captures(actor_id)
+          .map_or_else(String::new, |captures| format_apub_name::<PREFIX>(&captures[1], &captures[2]))
+      } else {
+        use js_sys::RegExp;
```
SleeplessOne1917, line 31: Using the browser's regex engine for WASM so we don't need to send a whole regex engine's worth of code in the WASM bundle.

src/utils/traits/to_str.rs
```diff
@@ -0,0 +1,13 @@
+pub trait ToStr {
+  fn to_str(self) -> &'static str;
+}
+
+impl ToStr for bool {
+  fn to_str(self) -> &'static str {
```
SleeplessOne1917, line 6: Small trait to remove the need to allocate a string on the heap when stringifying bools.

src/ui/components/layouts/base_layout/side_nav.rs
dessalines: Cool, the actual html tags matter to me less than just making sure the components are modular.

src/ui/components/common/content_actions/comment_count.rs
```diff
@@ -0,0 +1,25 @@
...
+  ui::components::common::icon::{Icon, IconType},
+  utils::types::Comments,
+};
+use leptos::*;
+use leptos_router::A;
+
+#[component]
+pub fn CommentCount(id: i32) -> impl IntoView {
+  let comments = expect_context::<Signal<Comments>>();
+  let num_comments_label = Signal::derive(move || format!("{} comments", comments.get().0));
```
dessalines, line 11: The comment count pry shouldn't be part of the context right, just a signal input. Also the param should pry be `post_id: PostId`, to be clear.

src/ui/components/common/content_actions/hide_post_button.rs
```diff
@@ -0,0 +1,34 @@
...
+
+  view! {
+    <li>
+      <ActionForm action=hide_post_action>
+        <input type="hidden" name="id" value=id />
+        <input type="hidden" name="hide" value=move || (!hidden.get().0).to_str() />
+        <button class="text-xs whitespace-nowrap" type="submit">
+          <Icon icon=icon class="inline-block" />
+          " "
+          {move || if hidden.get().0 { "Unhide post" } else { "Hide post" }}
```
dessalines, line 29: Not sure if you started i18n yet or not.

src/ui/components/common/content_actions/report_button.rs
```diff
@@ -0,0 +1,45 @@
+use crate::{
+  ui::components::common::icon::{Icon, IconType},
+  utils::types::{ContentActionType, ReportModalData, ReportModalNode},
+};
+use leptos::*;
+
+fn report_content(id: i32, content_type: ContentActionType, creator_actor_id: String) {
```
dessalines, line 7: This can be a `PostId` or `CommentId` right? It'd pry be best to do this either with an enum (I'm not sure if a simple union would work IE `PostId | CommentId`)

Something like 

```rust
enum PostOrCommentId {
  Post(PostId)
  Comment(CommentId)
}
```

src/ui/components/common/content_actions/report_button.rs
```diff
@@ -0,0 +1,45 @@
...
+  });
+  let _ = report_modal
+    .get_untracked()
+    .expect("Report dialog should exist")
+    .show_modal();
+}
+
+#[component]
+pub fn ReportButton(
+  id: i32,
```
dessalines, line 24: Same for all these IDs, make sure they're typed with `PostId`, `CommentId`, etc.

src/ui/components/common/content_actions/report_button.rs
```diff
@@ -0,0 +1,45 @@
...
+  content_action_type: ContentActionType,
+  creator_actor_id: StoredValue<String>,
+) -> impl IntoView {
+  let report_content_label = if content_action_type == ContentActionType::Comment {
+    "Report comment"
```
dessalines, line 29: possibly i18n

src/ui/components/common/creator_listing.rs
```diff
@@ -0,0 +1,33 @@
+use crate::utils::create_user_apub_name;
+use leptos::*;
+use leptos_router::A;
+
+#[component]
+pub fn CreatorListing(
+  avatar: Option<String>,
+  display_name: Option<String>,
+  name: String,
+  actor_id: String,
```
dessalines, line 10: You can probably pass the whole person. 

[Example from jerboa](https://github.com/LemmyNet/jerboa/blob/main/app/src/main/java/com/jerboa/ui/components/person/PersonProfileLink.kt#L66)

src/ui/components/modals/report_modal.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+use html::Dialog;
+use leptos::*;
+use leptos_router::ActionForm;
+
+#[component]
+fn ReportForm(
+  id: Signal<i32>,
+  creator_actor_id: Signal<String>,
+  content_action_type: Signal<ContentActionType>,
```
dessalines, line 20: Same ID stuff, you could use a PostOrComment enum to differentiate,

src/utils/apub_name.rs
dessalines: What's the purpose of this one? The actor ids should point to the correct locations already.

src/utils/types/dialog_types.rs
```diff
@@ -0,0 +1,12 @@
+use super::ContentActionType;
+use leptos::{html::Dialog, NodeRef};
+
+#[derive(Clone, Default)]
+pub struct ReportModalData {
+  pub id: i32,
+  pub content_type: ContentActionType,
+  pub creator_actor_id: String,
```
dessalines, line 8: Enum

dessalines (commented):
I tested and its looking really good so far. I have some suggestions below.

SleeplessOne1917:
https://github.com/LemmyNet/lemmy-ui-leptos/pull/107#discussion_r1694267404

> What's the purpose of this one? The actor ids should point to the correct locations already.

The actor ID is a URL to the user/comment, e.g. `https://instance.tld/u/name`. I need to do some parsing if I want to transform it into the stylized ID, e.g. `@name@instance.tld`. It's possible for me to pass the user/community name in, but sadly only the instance id is available. If I want to get the instance domain for the stylized apub links, the only place I could reliably get it was from the `actor_id`.

src/ui/components/common/community_listing.rs
```diff
@@ -0,0 +1,29 @@
...
+use leptos_router::A;
+
+#[component]
+pub fn CommunityListing(
+  icon: Option<String>,
+  title: String,
+  name: String,
+  actor_id: String,
+) -> impl IntoView {
+  view! {
```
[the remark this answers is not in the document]
SleeplessOne1917: I'm passing individual fields to avoid unnecessary cloning.

src/ui/components/common/creator_listing.rs
SleeplessOne1917: [Similar issue to here.](https://github.com/LemmyNet/lemmy-ui-leptos/pull/107#discussion_r1694356214)

src/ui/components/common/content_actions/hide_post_button.rs
SleeplessOne1917: I recall having trouble switching translations dynamically from user input. If we only want to use the language from the browser locale and not expose an option for users to change their interface language, we could probably go with i18n right away. Stilll have to check how it interacts with browser locale TBH.

src/ui/components/common/content_actions/report_button.rs
SleeplessOne1917: [Addressed in another comment](https://github.com/LemmyNet/lemmy-ui-leptos/pull/107#discussion_r1694356898).

src/utils/apub_name.rs
SleeplessOne1917: > What's the purpose of this one? The actor ids should point to the correct locations already.

The actor ID is a URL to the user/comment, e.g. `https://instance.tld/u/name`. I need to do some parsing if I want to transform it into the stylized ID, e.g. `@name@instance.tld`. It's possible for me to pass the user/community name in, but sadly only the instance id is available. If I want to get the instance domain for the stylized apub links, the only place I could reliably get it was from the `actor_id`.

src/ui/components/common/content_actions.rs
```diff
@@ -1,121 +1,135 @@
...
+pub fn ContentActions<SA>(
   content_action_type: ContentActionType,
-  #[prop(into)] id: MaybeSignal<i32>,
+  id: i32,
+  saved: Signal<bool>,
   save_action: ServerAction<SA>,
-  #[prop(into)] saved: MaybeSignal<bool>,
-  report_action: ServerAction<RA>,
-  #[prop(into)] creator_id: MaybeSignal<i32>,
+  creator_id: i32,
```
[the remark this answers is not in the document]
SleeplessOne1917: I'll have to take the tuple value out of the `PersonId` at some point. Doing so once in the post listing removes the need to do it multiple times in the content actions. I tried getting the best of both worlds by making the `creator_id` an into prop, but sadly `PersonId` doesn't implement `Into<i32>`.

src/ui/components/common/content_actions/comment_count.rs
SleeplessOne1917: I moved up the comment count to the post listing. This required that I move the wrapping `div` around the content actions up to the post listing, but this is a small inconvenience compared to using the context.

src/ui/components/common/content_actions/report_button.rs
SleeplessOne1917: I ended up making an enum like you suggested:

```rust
pub enum ContentId {
    Post(i32),
    Comment(i32)
}
```

[I used an `i32` for this for similar reasons I described in another comment.](https://github.com/LemmyNet/lemmy-ui-leptos/pull/107#discussion_r1694362491)

src/ui/components/common/content_actions/report_button.rs
SleeplessOne1917: [See my other relevent comment.](https://github.com/LemmyNet/lemmy-ui-leptos/pull/107#discussion_r1694362491)

src/ui/components/modals/report_modal.rs
SleeplessOne1917: Done

src/utils/types/dialog_types.rs
SleeplessOne1917: I don't think it's needed here. In this case, it's always going to be the id of a user.

src/utils/apub_name.rs
dessalines: Instead of regex, you can just get the hostname from the actor_id , then use the community / user name.

`@{name}@{hostName(person.actor_id)}` 

Anywhere you want to create these stylized names, you'll likely already have `Person` or `Community`, which has everything you need (name, display name, and actor_id).

Here's an [example for person and community](https://github.com/LemmyNet/jerboa/blob/main/app/src/main/java/com/jerboa/Utils.kt#L450)

src/ui/components/common/community_listing.rs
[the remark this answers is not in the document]
dessalines: A good rule of thumb for rust is to use references in function arguments. If they're not being mutated anyway then its usually fine, IE `fn CommunityListing(community: &Community) ...`

src/ui/components/common/content_actions.rs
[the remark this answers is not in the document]
dessalines: I'd need to get into the code and start playing, but I thought signals were generic, and those `#[prop(into)` macros are optional. So you could do `ReadSignal<PostId>` 

The only place I can think of where you'd absolutely need to take the tuple out, is in route params, and then you should probably wrap it in `PostId(` or whatever type to make sure it uses that ID type internally.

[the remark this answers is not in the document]
SleeplessOne1917: > I'd need to get into the code and start playing, but I thought signals were generic, and those #[prop(into) macros are optional. So you could do ReadSignal<PostId>

Since the id for a post, comment, or creator isn't something that can change (barring the report modal, since there's only one of them and it gets reused for each report), there's no need to make it reactive. In this case, a signal would just add extra overhead.

> The only place I can think of where you'd absolutely need to take the tuple out, is in route params, and then you should probably wrap it in PostId( or whatever type to make sure it uses that ID type internally.

It's also required to pass to the `value` attribute for hidden inputs in several places in the app.

At any rate, we'll need to take the tuple value out of the id at some point or another. As I stated above, doing so earlier rather than later reduces the number of times we need to do so. I can still change the components to use the `PostId`/`CommentId` if you still think it necessary in light of this information though.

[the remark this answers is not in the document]
dessalines: I do think its a good idea to keep those tuple types as much as possible. We could also probably add `Into<i32>` for those tuple structs in the lemmy codebase, but I'm not sure why that's necessary.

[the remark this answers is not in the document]
dessalines: Okay I opened up the code, and you should be fine using the `PersonId` or whichever internally, and then only extracting the inner when using it in html: ie  `value = {creator_id.0}`

src/utils/apub_name.rs
SleeplessOne1917: I'm using the `url` crate for the server-side and the browser URL binding for the client-side. It greatly simplified things.

dessalines:
I'll have a PR to this one shortly.

src/ui/components/common/community_listing.rs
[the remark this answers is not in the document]
SleeplessOne1917: I tried this approach and ran into a problem. Leptos components only seem to work with references when they are `'static`. I think it has something to do with Leptos relying very heavily on `move` closures. I'll ask in their Discord if anyone there has suggestions for avoiding so many clones.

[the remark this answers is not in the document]
dessalines: Hrm, that's surprising... I'll try this also.

dessalines (approved):

PR #111, opened 2024-07-29
Woodpecker update try 1

dessalines:
A few cleanups to the woodpecker, including:

- Not using the internally built docker images, but ones based off of rust images.
- Adding a few variables, and a binstall helper.


SleeplessOne1917:
Was there an issue with how it was working before? A big reason I built the images in steps in the Dockerfile was to avoid needing to do the same work repeatedly in the woodpecker CI. This solves the unnecessary duplication issue, but it still builds from scratch for the steps that used to use the `local/base` and `local/leptos-ui` images.

dessalines:
The major reason is that I couldn't do format-type checks before building those images, which wastes computing time.

I've added some build caching in between steps now via CARGO_HOME.

Also it doesn't seem too standard to be using internally built docker images, rather than just the rust images themselves. Until we figure out the proper way to do playwright checks in CI, we might as well just not build the images, seems like a waste.

.woodpecker.yml
```diff
@@ -1,5 +1,10 @@
 variables:
   - &cd_workdir "cd /usr/src/app"
+  - &rust_image "rustlang/rust:nightly-bookworm"
+  - install_binstall: &install_binstall
```
SleeplessOne1917, line 4: You could do what the Dockerfile is currently doing for this step:

```
wget -O- https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-x86_64-unknown-linux-musl.tgz | tar -xvz -C /usr/local/cargo/bin
```

This will also eliminate the need to copy `cargo-binstall` to `/usr/local/cargo/bin` since it will be extracted there directly.

.woodpecker.yml
```diff
@@ -17,99 +22,90 @@ steps:
...
-    detach: true
-    when:
-      - event: pull_request
-
-  lemmy:
-    image: dessalines/lemmy:0.19.4-beta.7
-    environment:
-      LEMMY_CONFIG_LOCATION: /woodpecker/src/github.com/LemmyNet/lemmy-ui-leptos/end2end/lemmy.hjson
-    detach: true
+  build_image:
```
SleeplessOne1917, line 86: Looks like this step isn't used anywhere anymore.

.woodpecker.yml
```diff
@@ -121,3 +117,20 @@ steps:
...
+    image: pgautoupgrade/pgautoupgrade:16-alpine
+    environment:
+      POSTGRES_USER: lemmy
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: lemmy
+    when:
+      - event: pull_request
+
+  lemmy:
+    image: dessalines/lemmy:0.19.4-beta.7
```
SleeplessOne1917, line 132: Can be 0.19.5

.woodpecker.yml
```diff
@@ -17,99 +22,90 @@ steps:
...
       - cargo leptos build
     when:
       - event: pull_request
-    depends_on: [build_image]
 
   # cargo_test:
   #   image: local/leptos-ui
   #   entrypoint: [cargo, leptos, test]
   #   when:
   #     - event: pull_request
```
SleeplessOne1917, line 84: Can be uncommented since there are tests for the apub name formatter now.

.woodpecker.yml
```diff
@@ -121,3 +117,20 @@ steps:
       - "curl -d'Lemmy-UI-leptos:${CI_COMMIT_TAG} deployed' ntfy.sh/lemmy_drone_ci"
     when:
       event: tag
+
+services:
```
SleeplessOne1917, line 121: Do these services get started up before any of the steps run?

SleeplessOne1917 (changes requested):

.woodpecker.yml
dessalines: Sweet, I'll try it out.

.woodpecker.yml
dessalines: Yep they do.

SleeplessOne1917 (approved):

PR #112, opened 2024-07-30
Using object references.

dessalines:
This should be a little cleaner. I'll comment on a few weird issues I had below.


Cargo.toml
```diff
@@ -70,6 +73,11 @@ ssr = [
   "dep:url",
 ]
 
+[lints.clippy]
+# leptos can't deal with these lints
+needless_lifetimes = "allow"
+let_unit_value = "allow"
```
dessalines, line 79: Leptos failed at ignoring these correctly.

src/ui/components/common/content_actions.rs
```diff
@@ -63,8 +60,11 @@ where
   });
   let crosspost_label = "Crosspost";
 
+  // TODO a very strange bug, where when using Show, you have to use a stored value
+  let creator_stored = StoredValue::new(creator.clone());
```
dessalines, line 64: Really weird one, and I wasted too much time on it. If the `<ReportButton` is outside of `<Show`, this isn't necessary. Inside show, it needs this specific weird cloning.

src/utils/apub_name.rs
```diff
@@ -14,14 +13,17 @@ fn format_apub_name<const PREFIX: char>(name: &str, instance: &str) -> String {
 }
 
 fn create_apub_name<const PREFIX: char>(name: &str, actor_id: &str) -> String {
+  // TODO Strange issue where the must be defined, or leptos will fail at startup
+  let default_url = "https://example.com";
```
dessalines, line 17: I started getting leptos startup errors all the sudden without this.

SleeplessOne1917: I made a slightly different fix for this issue in my fluent branch that I will be pushing up soon. I'll see if you like it more than this one.

src/ui/components/post/post_listing.rs
```diff
@@ -14,23 +14,14 @@ use leptos::*;
 use leptos_router::*;
 
 #[component]
-pub fn PostListing(post_view: PostView) -> impl IntoView {
+pub fn PostListing<'a>(post_view: &'a PostView) -> impl IntoView {
```
SleeplessOne1917, line 17: This is one area that I don't see the benefit of using a reference since the post view isn't reused in either the post listings or post page.

src/ui/components/common/content_actions/report_button.rs
```diff
@@ -24,27 +21,16 @@ fn report_content(
...
-  post_or_comment_id: PostOrCommentId,
-  creator_actor_id: StoredValue<String>,
-) -> impl IntoView {
+pub fn ReportButton<'a>(creator: &'a Person, post_or_comment_id: PostOrCommentId) -> impl IntoView {
   let report_content_label = if matches!(post_or_comment_id, PostOrCommentId::Comment(_)) {
     "Report comment"
   } else {
     "Report post"
   };
+  let onclick = report_content(creator, post_or_comment_id);
```
SleeplessOne1917, line 30: I was running into a bug when I tested your changes locally, and I believe the change to how the `onclick` function is handled is the cause. When I log in as a user, the report modal immediately pops up on the first post in the list. This doesn't happen in main. That's because this change makes it so that the `report_content` function is called when the component is rendered as opposed to when the button is clicked. If you want to move the `onclick` closure to a variable instead of inlining it when passed to the button, this sould be:

```rust
let onclick = move |_| report_content(creator, post_or_comment_id);
```

Then you can just pass it directly to the button:

```html
<button class="text-xs whitespace-nowrap" type="button" on:click=onclick>
```

src/ui/components/common/content_actions.rs
SleeplessOne1917: [This part of the Leptos book should make the issue a bit clearer.](https://book.leptos.dev/interlude_projecting_children.html)

Cargo.toml
```diff
@@ -35,6 +35,9 @@ tailwind_fuse = { version = "0.3", features = ["variant"] }
 web-sys = { version = "0.3", optional = true, features = ["Url"] }
 url = { version = "2.5", optional = true }
 
+[dev-dependencies]
+anyhow = "1.0.86"
```
SleeplessOne1917, line 39: I don't see this being used anywhere in the code. I don't think we need it at the moment.

SleeplessOne1917 (changes requested):
I'll give this another once over later, but at the very least the report content bug needs to be fixed.

src/ui/components/common/content_actions.rs
dessalines: Nice, thx.

src/ui/components/post/post_listing.rs
dessalines: It also prevents cloning wherever this is used.

src/ui/components/common/content_actions/report_button.rs
dessalines: Nice, done.

Cargo.toml
dessalines: I had used it in a test, but I'll remove it now.

SleeplessOne1917 (commented):
I'll test this locally once I'm at my computer. I won't be able to do that until late tonight or early tommorrow.

dessalines:
I couldn't get any dropdowns to work, even on the main branch, so I couldn't fully test.

src/cookie_middleware.rs
```diff
@@ -6,14 +6,20 @@ use actix_session::{
...
   .cookie_name(AUTH_COOKIE.into())
-  .cookie_secure(cfg!(debug_assertions))
+  .cookie_secure(!debug_mode)
   .session_lifecycle(PersistentSession::default())
-  .cookie_same_site(SameSite::Strict)
+  .cookie_same_site(if debug_mode {
+    SameSite::Lax
+  } else {
+    SameSite::Strict
+  })
```
SleeplessOne1917, line 22: Why make it lax in dev?

SleeplessOne1917:
I tested for the dropdown bug. It turns out it's not a bug, but is easily mistaken for one. Currently, the action dropdown doesn't list any actions users are able to perform on their own posts. If you try the dropdown on another user's posts, it should work. Relevant items will be added to the content dropdown in the near future.

SleeplessOne1917 (approved):

PR #123, opened 2024-08-07
Trying another SI number formatter.

dessalines:
Added a separate file, and a few tests.

Still not as good as lemmy-ui's [intl number format](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils/helpers/num-to-si.ts#L1), but it works.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ tokio = { version = "1.39", optional = true, features = ["macros"] }
 strum = { version = "0.26", features = ["derive"] }
 trait-set = "0.3"
 leptos-use = "0.11"
-si_format = "0.1"
+format_num = "0.1.0"
 tailwind_fuse = { version = "0.3", features = ["variant"] }
 web-sys = { version = "0.3", optional = true, features = ["Url"] }
 url = { version = "2.5", optional = true }
--- a/src/utils/format.rs
+++ b/src/utils/format.rs
@@ -1,0 +1,28 @@
+use format_num::NumberFormat;
+
+const PATTERN: &str = ".3s";
+
+pub fn si_format<T>(num: T) -> String
+where
+  T: Into<f64>,
+{
+  let fmt = NumberFormat::new();
+  fmt.format(PATTERN, num)
+}
+
+pub fn si_format_i64(num: i64) -> String {
+  si_format(num as f64)
+}
+
+#[cfg(test)]
+mod tests {
+  use crate::utils::format::si_format;
+
+  #[test]
+  fn format_check() {
+    assert_eq!(si_format(12e6), "12.0M");
+    assert_eq!(si_format(12.345), "12.3");
+    assert_eq!(si_format(12345), "12.3k");
+    assert_eq!(si_format(123456), "123k");
+  }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1218,6 +1218,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "format_num"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14ac05eb8d2eb4ed1eeff847911deae077b0b53332465de9d6a26b0ea9961bc8"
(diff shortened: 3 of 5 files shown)


src/utils/format.rs
```diff
@@ -0,0 +1,28 @@
...
+pub fn si_format_i64(num: i64) -> String {
+  si_format(num as f64)
+}
+
+#[cfg(test)]
+mod tests {
+  use crate::utils::format::si_format;
+
+  #[test]
+  fn format_check() {
```
SleeplessOne1917, line 22: How does this look when formatting single, double, or triple digit numbers, e.g. 5, 26, 123? Also, you could probably test using the i64 version.

src/utils/format.rs
```diff
@@ -0,0 +1,28 @@
+use format_num::NumberFormat;
+
+const PATTERN: &str = ".3s";
+
+pub fn si_format<T>(num: T) -> String
+where
+  T: Into<f64>,
+{
+  let fmt = NumberFormat::new();
```
SleeplessOne1917, line 9: Since this formatter is always the same, you could make this a static using `LazyLock`.

src/utils/format.rs
```diff
@@ -0,0 +1,28 @@
+use format_num::NumberFormat;
+
+const PATTERN: &str = ".3s";
+
+pub fn si_format<T>(num: T) -> String
```
SleeplessOne1917, line 5: This is only used in this module, so this shouldn't need to be public.

Cargo.toml
```diff
@@ -29,7 +29,7 @@ tokio = { version = "1.39", optional = true, features = ["macros"] }
 strum = { version = "0.26", features = ["derive"] }
 trait-set = "0.3"
 leptos-use = "0.11"
-si_format = "0.1"
+format_num = "0.1.0"
```
SleeplessOne1917, line 32: This library seems to get the job done. My only concern is that it hasn't been touched in over 4 years and was only made in 3 commits. At that point, I don't see the benefit to using this library when the hand-rolled implementation was less complex (we only use 1 pattern and this library uses regex to parse the pattern), and the hand-rolled implementation was thoroughly tested to make sure it formatted numbers in the way we expect it to.

src/utils/format.rs
dessalines: Looks like unfortunately it doesn't remove the extra decimals in a way that the `compact` setting of the `intl` does. IE so 5 -> `5.00`

Cargo.toml
dessalines: Yeah, doesn't look like the best library. I'll mark this as draft, until we find (or make) a better crate. 

This looks like a good simple possibility, I'll work on it soon: https://stackoverflow.com/a/23794345/1655478

SleeplessOne1917: [There's always the function from the old PR.](https://github.com/LemmyNet/lemmy-ui-leptos/blob/5b2e653467ee9ee34898b03de15b329a15faabb6/src/utils/format_number_si.rs) Not sure if it makes sense to extract into its own crate, but it's lightweight and behaves the way we expect it to.

dessalines: Ya I think it would be useful to extract that to its own crate. LMK if you want to, otherwise I can. Maybe `pretty_si_number_format` or something.

dessalines: Also do you remember where that algorithm came from?

SleeplessOne1917: I think it's fine to leave the implementation in this codebase for now. When I get some free time where I'm not working on this UI or the inferno one, I'll play around with adding some options to the algorithm. I think it's a bit too specific to our usecase to be extracted to its own crate for the time being.

As to where the algorithm came from, I came up with it myself.

dessalines: Mmmk, in that case I'll close this so you can just re-add that last one.

I'd pry encourage you to make the crate sooner rather than later, even if it only fits our one use-case, so others could potentially start contributing and adding the options they want themselves.

PR #124, opened 2024-08-10
Replace si_format with pretty-num

SleeplessOne1917:


dessalines (approved):
Nice, thx for making that.

PR #135, opened 2024-08-27
Some post related stuff

SleeplessOne1917:
This makes a few changes to make posts look right on both the post page and in post lists:

- Make posts look good on both mobile and larger screens
- Add post publish date and language
- Handle thumbnail display logic
- Tweak various styles, most notably adding a max width for content
- Refactor sidebar to make it easier to use for both instance information and community information

### Home page desktop/laptop
image

### Home page mobile
image

### Post page desktop/laptop
image

### Post page mobile
image

While this gets some things done, there are some notable omissions. The most obvious is the lack of post body - I will tackle that when adding markdown rendering. I also still need to find a way to display images in videos, whether that ends up being a popout like in the inferno UI or some other approach. I still need to add more content actions to the posts as well. Then there are comments, which is its own beast.


locales/en/main.ftl
```diff
@@ -67,4 +67,37 @@ admins = Admins
...
+downvote = Downvote
+local-subscribers = Local Subscribers
+subscribers = Subscribers
+community-stats = Community Stats
+moderators = Moderators
+now = Now
+years-ago = {$years ->
+    [one] {$years} year ago
+    *[other] {$years} years ago
+}
```
SleeplessOne1917, line 79: This style of singular or plural is where fluent is starting to show its edge over i18n. After looking at the fluent syntax guide, I discovered that [fluent supports number formatting that uses a similar API to browsers' `Intl.NumberFormat` class](https://projectfluent.org/fluent/guide/functions.html#number). It may even be possible to remove the dependency on the crate I created because of this, although I haven't tested this yet and will have to see if it behaves as expected.

src/ui/components/post/post_listing.rs
```diff
@@ -23,18 +32,53 @@ pub fn PostListing<'a>(post_view: &'a PostView) -> impl IntoView {
...
+    with!(
+      |post_state| post_state
+        .post
+        .thumbnail_url
+        .as_deref()
+        .map(AsRef::as_ref)
+        .map(Rc::from)
+        .or_else(|| post_url.get().filter(|url| is_image(url.as_ref()))) // Fall back to post url if no thumbnail, but only if it is an image url
+    )
+  });
```
SleeplessOne1917, line 54: I had to use `Rc`s for these because they wouldn't play nice with the compiler otherwise.

src/ui/components/post/post_listing/thumbnail.rs
SleeplessOne1917, line 1: Not the prettiest code, but it's much less of a clusterfuck than [the thumbnail logic in the inferno UI](https://github.com/LemmyNet/lemmy-ui/blob/2fdd42087aa7bc67e8a571e1cbcf630386133828/src/shared/components/post/post-listing.tsx#L306-L432).

src/utils/get_time_since.rs
```diff
@@ -0,0 +1,45 @@
...
+
+  let years = now.years_since(*date_time).unwrap_or_default();
+  if years > 0 {
+    return tr!("years-ago", { "years" => years });
+  }
+
+  let delta = now - date_time;
+
+  let weeks = delta.num_weeks();
+  let months = weeks / 4;
```
SleeplessOne1917, line 15: Sadly, there is no convenient method on `TimeDelta`s for getting the number of months, so this will have to suffice.

src/ui/components/common/vote_buttons.rs
```diff
@@ -41,6 +29,7 @@ pub fn VoteButtons<VA>(
   id: PostOrCommentId,
   score: Signal<i64>,
   vote_action: ServerAction<VA>,
+  #[prop(optional)] class: &'static str,
```
dessalines, line 32: I looked at [some of the docs for this](https://tailwind-fuse.vercel.app/), and it seems like these tailwind fuse classes are meant to be used to define various *styles*, not store extra data. 

So like `enum VoteStyle { None, FilledIn, Outlined }` or whatever it is.

This one also is confusing me because its an enum, but its being used like a component?

src/ui/components/post/post_listing.rs
dessalines: I'm trying to go through some of their logic here: 

- https://docs.rs/leptos/latest/leptos/fn.create_memo.html
- https://book.leptos.dev/reactivity/working_with_signals.html#making-signals-depend-on-each-other

Their main use case is for expensive computations, not simple reads, so I could see using one for the image_url (which has to do a regex check), but pry isn't necessary for the other simple reads. 

Also it seems like `create_memo(` is preferred over `Memo::new`

src/ui/components/post/post_listing/thumbnail.rs
dessalines: Looks much cleaner to me.

src/ui/components/post/post_listing/thumbnail.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[component]
+pub fn Thumbnail(
+  url: Memo<Option<Rc<str>>>,
+  image_url: Memo<Option<Rc<str>>>,
+  #[prop(into)] has_embed_url: Signal<bool>,
+  id: PostId,
+) -> impl IntoView {
+  let thumbnail_data = Signal::derive(move || {
```
dessalines, line 43: What's the point of using `Signal::derive` ? Seems like you can just do moves: https://book.leptos.dev/view/02_dynamic_attributes.html?highlight=derived#derived-signals

src/ui/components/post/post_listing/thumbnail.rs
```diff
@@ -0,0 +1,112 @@
...
+    with!(|url, image_url, has_embed_url| {
+      let (icon_size, icon_class, wrapper_class) = if image_url.is_some() {
```
dessalines, line 45: For all the if cases in this function, it'd be good to add a comment.

src/utils/filetype.rs
dessalines, line 1: Nice

src/utils/filetype.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub fn is_image(url: &str) -> bool {
+  is_filetype(url, &IMAGE_TYPES)
+}
+
+pub fn is_video(url: &str) -> bool {
+  is_filetype(url, &VIDEO_TYPES)
+}
+
+fn is_filetype(url: &str, exts: &[&str]) -> bool {
```
dessalines, line 12: Something tells me that using `path_segments.last()` from https://docs.rs/url/latest/url/ might be a little safer, but you've added enough tests here that this is pry fine.

src/utils/get_time_since.rs
```diff
@@ -0,0 +1,45 @@
+use chrono::{DateTime, Utc};
+use leptos_fluent::tr;
+
+pub fn get_time_since(date_time: &DateTime<Utc>) -> String {
+  let now = Utc::now();
+
+  let years = now.years_since(*date_time).unwrap_or_default();
+  if years > 0 {
+    return tr!("years-ago", { "years" => years });
```
dessalines, line 9: You can pry remove all these return statements, and the `;` 's

src/ui/components/post/post_listing.rs
SleeplessOne1917: > I'm trying to go through some of their logic here:
> https://docs.rs/leptos/latest/leptos/fn.create_memo.html
> https://book.leptos.dev/reactivity/working_with_signals.html#making-signals-depend-on-each-other
> Their main use case is for expensive computations, not simple reads, so I could see using one for the image_url (which has to do a regex check), but pry isn't necessary for the other simple reads.

In this case, I'm using `Memo` because if I use a regular derived signal, any change in `post_state` will cause reactive updates in every part of the UI that relies on parts of it regardless of if the relevant value has changed or not. Considering `PostView` is a large struct with many parts of the UI relying on only 1 or 2 of its properties, I think this choice is justified. [This section of the Leptos book talks about this behavior](https://book.leptos.dev/15_global_state.html?highlight=state#option-3-create-a-global-state-struct-and-slices):

> [B]ecause our whole state is wrapped in one signal, updating the value of one field will cause reactive updates in parts of the UI that only depend on the other... There’s a better way. You can take fine-grained, reactive slices by using [create_memo](https://docs.rs/leptos/latest/leptos/fn.create_memo.html) or [create_slice](https://docs.rs/leptos/latest/leptos/fn.create_slice.html) (which uses create_memo but also provides a setter). “Memoizing” a value means creating a new reactive value which will only update when it changes. “Memoizing a slice” means creating a new reactive value which will only update when some field of the state struct updates.

While that section is talking about global state for the entire app, the `post_state` is basically global state for a post listing.

> Also it seems like create_memo( is preferred over Memo::new

[The release for the most recent v0.7 beta suggests otherwise](https://github.com/leptos-rs/leptos/releases/tag/v0.7.0-beta):
> We're migrating away from create_ naming toward more idiomatic Rust naming patterns:
> - create_signal to signal (like channel)
> - create_rw_signal to RwSignal::new()
> - etc.
> - I've left some of the current functions in, marked deprecated; others may have been missed, but should be easy to find [via docs.rs](https://docs.rs/leptos/0.7.0-alpha/leptos).

src/ui/components/common/sidebar.rs
```diff
@@ -0,0 +1,136 @@
...
+  view! {
+    <div class="card w-full mb-3 bg-base-200">
+      <div class="card-body">
+        <Transition>
+          <Unpack item=data let:data>
+            <h2 class="card-title">{data.name()}</h2>
+            {data.description().map(|description| view! { <p>{description}</p> })}
+            <section aria-labelledby="instance-stats-heading" class="my-4">
+              <h3 id="instance-stats-heading" class="text-2xl font-bold mb-2">
+                {if matches!(data, SidebarData::Site { .. }) {
```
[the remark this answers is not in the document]
SleeplessOne1917: In cases where a conditional only has 2 branches, I find `if/else` easier to read thatn`match`. However, setting a bunch at once does sound a bit easier and if `match` ends up being more readable in that case, I'll use that.

src/ui/components/common/sidebar.rs
```diff
@@ -0,0 +1,136 @@
...
+                          }
+                              .into_view()
+                      }
+                  }}
+                </tbody>
+              </table>
+            </section>
+            <section aria-labelledby="instances-admins-heading">
+              <h3 id="instance-admins-heading" class="text-2xl font-bold mb-2">
+                {if matches!(data, SidebarData::Site { .. }) {
```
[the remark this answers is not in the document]
SleeplessOne1917: [Addressed here.](https://github.com/LemmyNet/lemmy-ui-leptos/pull/135#discussion_r1734880653)

src/ui/components/common/sidebar/user_stat_row.rs
```diff
@@ -0,0 +1,12 @@
+use leptos::*;
+use pretty_num::PrettyNumber;
+
+#[component]
+pub fn UserStatRow(count: i64, text: Signal<String>) -> impl IntoView {
```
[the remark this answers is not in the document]
SleeplessOne1917: `Signal` is also read-only. Since this component is only meant to be used in the context of the sidebar and it always uses the return value of `move_tr!`, which is `Signal<String>`, I think it's a bit simpler to just use `Signal` here.

Also worth noting that `ReadSignal`s have no way of creating one directly like can be done with `Signal::derive`; rather, they are returned from either `create_signal` (as the first value in the tuple it returns), `RwSignal::read_only`, or `RwSignal::split` (again, as the first value in the return tuple similar to `create_signal`.

src/ui/components/common/vote_buttons.rs
SleeplessOne1917: > ...and it seems like these tailwind fuse classes are meant to be used to define various styles, not store extra data.

That's what I'm using it for, although it can be a bit difficult to grasp at first. I also struggled with it when I first used the library. I think, in the context of the vote buttons, it becomes clearer when you look at how it's used in this component.

Looking at the upvote button part of the `VoteButtons` component, we see this:

```rust
<button
    type="submit"
     class=move || {
         with!(
             | is_upvote | if * is_upvote { Vote::Up } else { Vote::None }.as_class().to_owned()
        )
    }

    title=move_tr!("upvote")
    disabled=move || !user_is_logged_in.get() || vote_action.pending().get()
>

    <Icon icon=IconType::Upvote />
</button>
```

Notice that `as_class` is called on the result of the conditional for `class` (which is a `VoteEnum` value). If this was expressed with using a string directly, it would look something like the following:

```rust
class=move || {
    with!(
        |is_upvote| tw_join!("align-bottom disabled:cursor-not-allowed disabled:text-neutral-content", if *is_upvote {"text-success"} else {"text-neutral"})
    )
}
```

I suppose those styles could be extracted to consts though, so it could end up looking something like this:

```rust
class=move || {
    with!(
        |is_upvote| tw_join!(vote_base_class, if *is_upvote {vote_success_class} else {no_vote_class})
    )
}
```

However, looking at it now, the `tw_join!` looks cleaner in this scenario, and the tailwind fuse enums/types seem to be best when they're being used as arguments. Alternatively, I could also try extracting the individual buttons into components, which could simplify the `VoteButtons` component and make the enums more useful. I'll play around with it a bit and see what looks better.

src/ui/components/post/post_listing/thumbnail.rs
SleeplessOne1917: You're probably right in this situation since I don't need to worry about passing it as an argument and needing to fiddle with generics.

src/utils/filetype.rs
SleeplessOne1917: Creating a `Url` with the `url` crate allocates the entire url as a string on the heap (as would calling `to_lowerrcase` on the `url` param), whereas this approach only allocates a `String` for the file extension of the url. I'm arguably being too anal about performance here, but I personally don't find the code too difficult to understand, its use is isolated to only this function as the single source of truth, and, as you pointed out, it has good test coverage.

src/utils/get_time_since.rs
SleeplessOne1917: I gave it a try and it didn't work. I think that approach would only work if the conditionals were all part of one big `if/else if/else` expression.

src/ui/components/common/vote_buttons.rs
SleeplessOne1917: I just refactored the vote buttons. Take a look and see if you like it.

src/ui/components/post/post_listing.rs
dessalines: You're ahead of the docs, nice.

src/ui/components/common/sidebar/user_stat_row.rs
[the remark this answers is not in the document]
dessalines: Ah, so move_tr requires it, gotcha.

src/ui/components/common/vote_buttons.rs
dessalines: Looks much better.

src/utils/filetype.rs
dessalines: Oh yeah I forgot about that... yep this is fine then.

src/ui/components/post/post_listing/thumbnail.rs
SleeplessOne1917: Added coments.

src/ui/components/common/sidebar/team_member_card.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[component]
+pub fn TeamMemberCard(person: Person) -> impl IntoView {
+  view! {
+    <li class="flex-1 text-center max-w-fit rounded-lg p-3 even:bg-base-100 odd:bg-base-300 shadow-lg shadow-neutral">
+      <img
+        src=person
+            .avatar
+            .map(|p| p.to_string())
+            .unwrap_or(String::from("/assets/default-avatar.png"))
```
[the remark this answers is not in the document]
SleeplessOne1917: I made a const for the avatar URL. The default community icon isn't used anywhere else yet, so I'm not sure it's worth extracting that at the moment. Not sure what you mean by "missings".

SleeplessOne1917:
All I have left before all comments are addressed is refactoring the sidebar logic.

SleeplessOne1917:
@dessalines I tried to implement some of the sidebar refactors you had in mind and they ended up being trickier than first anticipated. [The `SideBarExtras` approach](https://github.com/LemmyNet/lemmy-ui-leptos/pull/135#discussion_r1734728210) ended up adding another layer of indirection that made things harder to maintain. To use some pseudo code:

```rust
view! {
    // Assume this is in a component using the suggestion

    <Transition>
        <Unpack item=sidebar_data let:sidebar_data>
            {match sidebar_data {
                SiteSidebar(site_sidebar) => view!{
                    <h2>{site_sidebar.name()}</h2>
                    <p>{site_sidebar.description()}</p>
                     <div>
                         <div>Today: {site_sidebar.users_today()}</div>
                         <div>Past week: {site_sidebar.users_week()}</div>
                     </div>
                     // Do more stuff with site sidebar data
                },
               CommunitySidebar(community_sidebar) => view!{
                   <h2>{community_sidebar.name()}</h2>
                   <h3>{community_sidebar.title()}</h3> // Info unique to sidebar
                   <p>{community_sidebar.description()}</p>
                   // Carries on in a way similar to the site sidebar above
               }
            }}
        </Unpack>
    </Transition>
}
```

As can be seen above, since the `SidebarExtras` trait is implemented in data stored as values inside the `SidebarData` enum instead of implemented on the enum itself, this means we need to match use the functions from the trait, leading to a lot of duplication. The current approach allows information that is common to both sites and communities to be used and only requires matching when using information specific to one of them.

Regarding [moving the uses of the `matches!` macro outside of the view so mutiple can be used at once](https://github.com/LemmyNet/lemmy-ui-leptos/pull/135#discussion_r1734880653): the type of the `data` prop for `Sidebar` is a `QuerySignal<SidebarData>` (a type alias for `Signal<Option<Result<SidebarData, ServerFnError>>>`), which makes it very difficult to use outside of an `<Unpack>` component. I could try my `derive_query_signal` function, but it would also require adding in more `<Unpack>`s and I'm not sure if it's worth it. I really hope Leptos 0.7 makes it easier to work with data for suspenses, since that difficulty is the reason I made the `<Unpack>` component in the first place.

I could pull the trigger on making `SidebarData` store a `Site` and a `Community` in its variants if you think that's important. While I do prefer to avoid cloning data I don't need to clone (both of those structs have a decent number of non-`Copy` fields), I already go against this concern by having the `team` prop accept a `QuerySignal<Vec<Person>>`. This issue I think ultimately stems from the sheer amount of superfluous data that we return from the API for many of our endpoints. I know we already have a discussion about this in the [API v4 backend issue](https://github.com/LemmyNet/lemmy/issues/4428), so I won't go on more about it here.

At any rate, I'll definitely need to touch the sidebar stuff again in the future when working on this project, so I'm sure this isn't the last time we'll be looking to refactor this.

dessalines (approved):
Mmmk, I'll try a pass at the enum / trait thing, but it a different branch. You can merge this whenever you like.

PR #136, opened 2024-08-29
Cleaning up sidebar.

dessalines:
I'll leave some comments below.

### DIFF
--- a/src/ui/components/common/sidebar.rs
+++ b/src/ui/components/common/sidebar.rs
@@ -1,135 +1,152 @@
-use crate::{
-  ui::components::common::{
-    icon::{Icon, IconSize, IconType},
-    unpack::Unpack,
-  },
-  utils::types::QuerySignal,
-};
-use lemmy_client::lemmy_api_common::lemmy_db_schema::source::person::Person;
+use crate::ui::components::common::icon::{Icon, IconSize, IconType};
 use leptos::*;
 use leptos_fluent::move_tr;
 use pretty_num::PrettyNumber;
+use sidebar_data::SidebarData;
 use team_member_card::TeamMemberCard;
 use user_stat_row::UserStatRow;
 
-mod sidebar_data;
-pub use sidebar_data::SidebarData;
+pub mod sidebar_data;
 mod team_member_card;
 mod user_stat_row;
 
 #[component]
-pub fn Sidebar(data: QuerySignal<SidebarData>, team: QuerySignal<Vec<Person>>) -> impl IntoView {
+pub fn Sidebar<'a>(data: &'a SidebarData) -> impl IntoView {
   let today = move_tr!("today");
   let past_week = move_tr!("past-week");
   let past_month = move_tr!("past-month");
   let past_6_months = move_tr!("past-6-months");
   let time_frame = move_tr!("time-frame");
 
+  // Extract the common elements
+  let (
+    heading,
+    team_heading,
+    team,
+    name,
+    description,
+    posts,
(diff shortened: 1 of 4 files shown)


src/ui/components/common/sidebar.rs
```diff
@@ -1,135 +1,152 @@
...
 
-mod sidebar_data;
-pub use sidebar_data::SidebarData;
+pub mod sidebar_data;
 mod team_member_card;
 mod user_stat_row;
 
 #[component]
-pub fn Sidebar(data: QuerySignal<SidebarData>, team: QuerySignal<Vec<Person>>) -> impl IntoView {
+pub fn Sidebar<'a>(data: &'a SidebarData) -> impl IntoView {
```
dessalines, line 14: Always try to pass the simple objects into components. The Top-level "screens" or "pages" should be doing all the work of extracting API responses, and components should be taking in simple data objects.

src/ui/components/common/sidebar.rs
```diff
@@ -1,135 +1,152 @@
...
   let today = move_tr!("today");
   let past_week = move_tr!("past-week");
   let past_month = move_tr!("past-month");
   let past_6_months = move_tr!("past-6-months");
   let time_frame = move_tr!("time-frame");
 
+  // Extract the common elements
+  let (
```
dessalines, line 22: The trait logic wasn't necessary anyway, since we still have to do a match for the common elements.

src/ui/components/common/sidebar/sidebar_data.rs
```diff
@@ -1,83 +1,24 @@
...
-    #[allow(dead_code)]
-    title: String,
-    #[allow(dead_code)]
-    icon: Option<String>,
-    description: Option<String>,
-    counts: CommunityAggregates,
-  },
+pub struct SiteSidebarData {
+  pub site: Site,
+  pub admins: Vec<Person>,
```
dessalines, line 9: I moved the admins / moderators into the sidebar data.

src/ui/components/common/sidebar.rs
```diff
@@ -1,135 +1,152 @@
...
+              <span class="text-sm">{move_tr!("posts")}</span>
+            </div>
+            <div>
+              <Icon icon=IconType::Comments size=IconSize::Large class="inline" />
+              {comments.pretty_format()}
+              " "
+              <span class="text-sm">{move_tr!("comments")}</span>
+            </div>
+
+            {if let SidebarData::Site(s) = data {
```
dessalines, line 86: We should probably even make an internal rule, that all views need their own components. It might seem verbose, but then we'd avoid all the mess of massive components like we have with lemmy-ui. 

We'd also get the benefit of seeing a named function that tells you what this block does.

SleeplessOne1917: > We should probably even make an internal rule, that all views need their own components.

Is a view here referring to a page? And by own components, do you mean just that there is a component for each (like is the case now), or that all pages should have their own components that are specific to them?

> It might seem verbose, but then we'd avoid all the mess of massive components like we have with lemmy-ui.

Breaking repeating things into smaller components is also an option.

SleeplessOne1917 (approved):
Looks good to me.

src/ui/components/common/sidebar.rs
dessalines: I just mean if you need to embed a view macro within a view macro, then its probably a good indicator that the inner view, should be its own component.

PR #143, opened 2024-09-20
Scroll tweaks

SleeplessOne1917:
Tweaks how scrolling is handled somewhat. Before, the only way to scroll posts was to scroll inside the post listing: if you were using a screen large enough to have margin between the edges of the viewport and the part of the screen with posts, you would get the aggravating experience of not being able to scroll. This was particularly noticeable and annoying if you were trying to scroll on the page for a single post. Now, you can scroll up and sown as long as you are not in the side nav.

Also, the sidebar is no longer independently scrollable. It still sticks to the side of the screen like it did before.

## Before
[Screencast_20240919_214514.webm](https://github.com/user-attachments/assets/b6e86f7d-50f3-44af-a1e0-ddb6239286c3)

## After
[Screencast_20240919_214258.webm](https://github.com/user-attachments/assets/dc1ad3e0-fb48-46ac-91b3-2267cdaaaffc)

### DIFF
--- a/src/ui/components/post/post_listings.rs
+++ b/src/ui/components/post/post_listings.rs
@@ -5,7 +5,7 @@ use leptos::*;
 #[component]
 pub fn PostListings(#[prop(into)] posts: MaybeSignal<Vec<PostView>>) -> impl IntoView {
   view! {
-    <ul class="divide-y divide-neutral sm:overflow-y-auto min-h-0 h-fit sm:h-full mb-14 sm:mb-4">
+    <ul class="divide-y divide-neutral min-h-0 h-fit sm:h-full mb-14 sm:mb-4">
       <For each=move || posts.get() key=|pv| pv.post.id let:pv>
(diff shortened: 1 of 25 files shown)


src/ui/components/post/post_listings.rs
```diff
@@ -5,7 +5,7 @@ use leptos::*;
 #[component]
 pub fn PostListings(#[prop(into)] posts: MaybeSignal<Vec<PostView>>) -> impl IntoView {
   view! {
-    <ul class="divide-y divide-neutral sm:overflow-y-auto min-h-0 h-fit sm:h-full mb-14 sm:mb-4">
+    <ul class="divide-y divide-neutral min-h-0 h-fit sm:h-full mb-14 sm:mb-4">
```
dessalines, line 8: So main change was just moving these overflows out into the main page then, gotcha.

dessalines (approved):

PR #186, opened 2024-11-11
Markdown Parsing

SleeplessOne1917:
I use the body of [this post showing off the markdown features](https://hexbear.net/post/3315660) to test.

# Headings (and monospace)
image

# Block Quotes
image

# Lists
image

# Image/emoji
image

# Subscript/superscript/ruby
image

# Tables
image

# Spoiler
image
image

Currently inline footnotes and code blocks are still not there. The former is due to a conflict between the footnote plugin and my superscript plugin. [I opened an issue a while back to work with the maintainer, but got no response.](https://github.com/markdown-it-rust/markdown-it-plugins.rs/issues/23) The latter I think is worth deferring until we have a solid idea of how we're handling themes, since `syntect` (which `markdown-it` uses for highlighted code blocks) has its own way of theming.


style/tailwind.css
```diff
@@ -1,3 +1,141 @@
...
+  @apply text-2xl font-black;
+}
+
+.markdown-content h2 {
+  @apply text-2xl font-bold;
+}
+
+.markdown-content h3 {
+  @apply text-xl font-black;
+}
```
SleeplessOne1917, line 19: Trying to do the styling with tailwind classes got pretty ugly since I can't add the classes directly to the elements rendered from the markdown, so I did the styling here while using tailwind classes for everything that wasn't a one-off style.

public/icons.svg
SleeplessOne1917, line 1: Made some changes to this so I could use SVG sprites. [This StackOverflow thread explains it well.](https://stackoverflow.com/questions/77133659/how-do-i-use-the-icon-from-the-svg-sprite-directly-in-the-css-scss)

style/tailwind.css
```diff
@@ -1,3 +1,141 @@
...
+.markdown-content blockquote em {
+  @apply not-italic;
+}
+
+.markdown-content blockquote:not(blockquote blockquote)::after {
+  content: " ";
+  mask-image: url("/icons.svg#css-quote");
+  -webkit-mask-image: url("/icons.svg#css-quote");
+  @apply bg-info block size-6 absolute end-1 bottom-1;
+}
```
SleeplessOne1917, line 58: The `mask-image` approach lets me control the color of the icon being used from a CSS URL.

style/tailwind.css
dessalines: No probs, ya since you're using tailwind classes anyway then we'll probably be fine.

style/tailwind.css
```diff
@@ -1,3 +1,141 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+.markdown-content {
+  @apply bg-base-200 mt-4 p-5 rounded space-y-4;
+}
+
+.markdown-content h1 {
+  @apply text-2xl font-black;
```
dessalines, line 10: `font-black` should probably be removed, as it might not work well with a dark theme of daisy-ui.

dessalines (approved):
Looks really good 👍 . Merge whenever you see fit.

style/tailwind.css
SleeplessOne1917: Despite the name, `font-black` is setting the font weight, not the text color. https://tailwindcss.com/docs/font-weight

dessalines: Ah cool.

PR #252, opened 2025-01-30
Update to use Leptos 0.7

SleeplessOne1917:
To test it locally, I used a copy of Lemmy's backend repo, checked out the 0.19.8 tag, and ran `docker_update.sh`. Then, with the backend handled, this can be run normally with `cargo leptos watch` (don't forget to use `pnpm install` first).


src/ui/components/common/filter_bar.rs
```diff
@@ -0,0 +1,43 @@
...
+
+mod listing_type_link;
+mod sort_type_link;
+
+#[component]
+pub fn FilterBar() -> impl IntoView {
+  view! {
+    <div class="mb-4 flex flex-wrap gap-3">
+      <div class="join">
+        <button class="btn join-item btn-active">{move_tr!("posts")}</button>
```
dessalines, line 16: Random question, do these strings get compile time checked?

src/ui/components/home/home_page.rs
```diff
@@ -1,77 +1,57 @@
...
+          <Transition>
+            <FilterBar />
+          </Transition>
         </div>
         <Suspense fallback=|| move_tr!("loading")>
           <ErrorBoundary fallback=|_| { move_tr!("could-not-load-posts") }>
-            <Unpack item=posts let:posts>
-              <PostListings posts=posts />
-            </Unpack>
+            <PostListings />
```
dessalines, line 33: It would probably be better to have the fetching take place outside, and postlistings take posts as a prop. You'll have to re-use it for the community page anyway. Not a blocker of course just something to keep in mind.

src/ui/components/home/home_page.rs
```diff
@@ -1,77 +1,57 @@
...
         </Suspense>
+
       </main>
 
       <aside class="hidden basis-7/20 xl:basis-3/10 lg:block me-8 sticky top-6 h-fit">
         <Transition>
-          <Unpack item=sidebar_data let:data>
-            <Sidebar data=&data />
-          </Unpack>
+          {move || Suspend::new(async move {
```
dessalines, line 41: If possible, its usually a good idea to try to move the fetching code above the UI code, near the top of the component.

src/ui/components/post/post_listing/thumbnail.rs
```diff
@@ -29,92 +29,104 @@ enum ThumbnailWrapperType {
...
-  pub icon_class: Rc<str>,
+  pub wrapper_class: Arc<str>,
+  pub icon_class: Arc<str>,
 }
 
 #[component]
 pub fn Thumbnail(
-  url: Memo<Option<Rc<str>>>,
-  image_url: Memo<Option<Rc<str>>>,
+  url: Memo<Option<Arc<str>>>,
```
dessalines, line 38: Is `Arc` supposed to be used for signals / memos?

dessalines (approved):
Thx, it'll be nice when leptos slows down a bit so these upgrades don't require so much time. 

BTW you can replace cargo machete with cargo shear [like we've done in lemmy](https://github.com/LemmyNet/lemmy/blob/main/.woodpecker.yml#L80), but you still might need to add some ignores to pass CI.

src/ui/components/common/filter_bar.rs
SleeplessOne1917: If you mean checking for whether or not a class is a valid Tailwind CSS class, no. Not quite sure what you mean though.

src/ui/components/post/post_listing/thumbnail.rs
SleeplessOne1917: See [the Leptos v0.7.0 release notes](https://github.com/leptos-rs/leptos/releases/tag/v0.7.0) section on Send/Sync Signals. Basically, signals need to be threadsafe now, and since this was already using the non-threadsafe `Rc`, `Arc` was the natural choice since it lets us do everything `Rc` does but threadsafe.

src/ui/components/home/home_page.rs
SleeplessOne1917: I originally moved it there becasue I mistook the false-positive warning from having one resource depend on another resource as an actual error. I hoisted it up a level.

PR #287, opened 2025-08-22
Make tailwind.config.js optional in Dockerfile

jeff-johnson-tech:


dessalines (commented):
Weird syntax for this. Anyway what is this for?

jeff-johnson-tech:
The Docker COPY command exits and fails the Docker build when the file (i.e. tailwind.config.js) doesn't exist. This fixes it by using a go filepath pattern which doesn't fail the build when the file doesn't exist. It was done this way to avoid potential regressions, but thinking back it's an over-engineered solution.

The weird syntax makes the file optional:
https://stackoverflow.com/questions/31528384/conditional-copy-add-in-dockerfile#answer-31532674

It is simpler to remove tailwind.config.js from the Dockerfile, since it's optional in Tailwind CSS v4.0 and likely not needed.

PR #301, opened 2025-10-12
Upgrading leptos.

dessalines:
I'm trying to upgrade leptos to the newest version, but I'm getting the following errors for every server fn:

```rust
error: future cannot be sent between threads safely
  --> src/serverfns/auth/login.rs:4:1
   |
 4 | #[server(prefix = "/serverfn")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ future created by async block is not `Send`
   |
   = help: within `{async block@src/serverfns/auth/login.rs:4:1: 4:32}`, the trait `Send` is not implemented for `Rc<RefCell<actix_session::session::SessionInner>>`
note: future is not `Send` as this value is used across an await
  --> src/serverfns/auth/login.rs:17:40
   |
 9 |   let (client, session) = get_client_and_session().await?;
   |                ------- has type `actix_session::Session` which is not `Send`
...
17 |   if let Some(jwt) = client.login(req).await.map_err(ServerFnError::new)?.jwt {
   |                                        ^^^^^ await occurs here, with `session` maybe used later
note: required by a bound in an opaque type
  --> src/serverfns/auth/login.rs:4:1
   |
 4 | #[server(prefix = "/serverfn")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: this error originates in the attribute macro `server` (in Nightly builds, run with -Z macro-backtrace for more info)

```

I

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,7 +8,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5f7b0a21988c1bf877cf4759ef5ddaac04c1c9fe808c9142ecb78ba97d97a28a"
 dependencies = [
- "bitflags 2.9.0",
+ "bitflags",
  "bytes",
  "futures-core",
  "futures-sink",
@@ -21,17 +21,17 @@ dependencies = [
 
 [[package]]
 name = "actix-files"
-version = "0.6.6"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0773d59061dedb49a8aed04c67291b9d8cf2fe0b60130a381aab53c6dd86e9be"
+checksum = "6c0d87f10d70e2948ad40e8edea79c8e77c6c66e0250a4c1f09b690465199576"
 dependencies = [
  "actix-http",
  "actix-service",
  "actix-utils",
  "actix-web",
- "bitflags 2.9.0",
+ "bitflags",
  "bytes",
- "derive_more 0.99.19",
+ "derive_more 2.0.1",
  "futures-core",
  "http-range",
  "log",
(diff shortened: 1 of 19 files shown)


dessalines:
I'm going over this, and action forms seem to make it impossible to push up events to screens, rather than making the API calls from within components(which can lead to spaghetti code) I might try to mess with ripping out the serverfns entirely, as it doesn't seem too necessary.

SleeplessOne1917:
I was planning on doing a similar Leptos upgrade with the branch updating to using v1 types.

Regarding the `ActionForm`s, the main benefit is that they provide a way to make the action work even when JS is disabled while still providing a smooth experience when JS is enabled. While I'll have to refresh myself on Leptos's newest versions, I recall that signals were changed in a way related to multithreading. Regardless, since the requests to server functions in action forms get their values from HTML input fields, it shouldn't result in API logic being deep in components. Also keep in mind that contexts in Leptos are documented (unlike in Inferno), so it's possible to keep some things top level and use them deeper in the component tree without prop-drilling.

dessalines:
IMO its probably not worth it to try to support non-JS actions. Maybe read-only is fine, but the extra complication that server functions and action forms adds is probably not worth it.

I created a [simple events example in leptos here](https://github.com/dessalines/leptos-events-testing/blob/main/src/main.rs#L19) that we should try to follow. 

It goes by [the jetpack compose declarative UI standard](https://developer.android.com/develop/ui/compose/mental-model), and its been a really good UI pattern to follow. You only push down data, and push up events. Never making API or server calls within any function except your top-level screens.

As far as contexts, IMO we should avoid them just like global statics. Prop drilling can be annoying, but its also really clean, preserves types, and every component has everything it needs, and can be tested in isolation. When adding a param, you'll be able to trace down every specific function that's missing this input, which isn't something you have with using global contexts (and I imagine they also can't deal with types). I don't even know how recomposition would properly work with changing globals.

I'll make an issue for this, but I'd like to make component "Preview" pages, where we're able to set up some sample data, then display all the components we're working on. IE a `/post/preview` page would display and display every component we're working on, so we can see them in absence of any API or the rest of the site setup.

SleeplessOne1917:
While I disagree regarding server functions, I'll defer to your judgement since you are significantly more involved in maintaining Lemmy than I am and this UI project has been in stasis for a long time. The progressive enhancement/graceful degredation benefit could be added much later (if at all) once the UI is production ready and the backend isn't undergoing so many changes.

There are a couple of points about contexts I should bring up though. They don't have to be (and often aren't) global throughout the whole component tree. They can be provided at any point in the component tree and can be retrieved in any component that is a descendant of that component. WRT types, Leptos actually retrieves a value from a context using its type. For example, getting a signal with a string from a context would look like this:

```rust
let value = use_context::<Signal<String>>().expect("Value not there");
```

The use of types this way discourages overusing contexts in situations they're not appropriate.

I'll mark this as approved once the errors are gone.

dessalines:
Might be a bit before I can finish this, as I realize I gotta focus more on lemmy-ui cause there's so much work we need to do there. I'd like to work on both, because this repo should be lemmy's main future UI, but we need like 3 more devs.

